PR #7, opened 2022-03-10
Add aggregates client

morningvera:
Defines a reusable client type and adds models and a client for our aggs API. Let's use this PR as a chance to iron out naming conventions, the general look of the public interface, and other patterns.


go.mod
```diff
@@ -1,3 +1,29 @@
 module github.com/polygon-io/client-golang
```
mcdayoub, line 1: We plan on changing this to `client-go`

mcdayoub: As well as rename the repo to `client-go`

rest/aggregates/aggregates.go
```diff
@@ -0,0 +1,45 @@
...
+	GetPreviousClosePath  = "/v2/aggs/ticker/{ticker}/prev"
+	GetGroupedDailyPath   = "/v2/aggs/grouped/locale/{locale}/market/{marketType}/{date}"
+	GetDailyOpenClosePath = "/v1/open-close/{ticker}/{date}"
+)
+
+type Client struct {
+	client.HTTPBase
+}
+
+func (ac *Client) Get(ctx context.Context, params GetParams, opts ...client.Option) (*AggsResponse, error) {
```
morningvera, line 23: We're leaving out the resource type in method names since users will call these methods with the package name as a prefix (e.g. `aggregates.Get(...)`). Open to changing it back to our current pattern though (`GetAggs`).

rest/aggregates/models.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	if !p.QueryParams.Adjusted {
+		q["adjusted"] = "false"
+	}
+
+	return q
+}
+
+type DailyOpenCloseResponse struct {
```
mcdayoub, line 164: The endpoint for `Daily Open/Close` provides a different response than the other Aggregates endpoints.

Also, the crypto-specific Daily Open/Close endpoint has a different set of path params and a different response.
https://polygon.io/docs/crypto/get_v1_open-close_crypto__from___to___date

Just something to keep in mind.

rest/aggregates/aggregates.go
```diff
@@ -0,0 +1,45 @@
+package aggregates
```
morningvera, line 1: is `aggs` a better package name?

rest/client/client.go
```diff
@@ -0,0 +1,211 @@
...
+
+	return statusErr
+}
+
+func (e *ErrorResponse) Error() string {
+	return fmt.Sprintf("bad status with code: %d; message: %s; request ID: %s; internal status: %s", e.StatusCode, e.Message, e.RequestID, e.Status)
+}
+
+// Options are used to configure client calls.
+type Options struct {
```
mcdayoub, line 150: Calling this `Options` is interesting because it could create confusion with our Options Market Type

rest/client/client.go
```diff
@@ -0,0 +1,211 @@
...
+	// If set, the RequestID will be passed downstream, otherwise the downstream service will create its own RequestID
```
qrpike, line 151: Is there a potential issue with people being able to set their own request IDs? Collisions, max character lengths, etc.

rest/client/client.go
```diff
@@ -0,0 +1,211 @@
...
+
+// HTTPBase provides functionality to make API requests via HTTP.
+type HTTPBase struct {
+	rc *resty.Client
+}
+
+type HTTPBaseConfig struct {
+	URL        string
+	Key        string
+	MaxRetries int
```
qrpike, line 35: Is this the default max retries, but the user is able to adjust this on a per call basis?

rest/aggregates/aggregates.go
```diff
@@ -0,0 +1,45 @@
...
+	client.HTTPBase
+}
+
+func (ac *Client) Get(ctx context.Context, params GetParams, opts ...client.Option) (*AggsResponse, error) {
+	res := &AggsResponse{}
+	err := ac.Call(http.MethodGet, GetPath, params, res, append([]client.Option{client.WithContext(ctx)}, opts...)...)
+	return res, err
+}
+
+func (ac *Client) GetPreviousClose(ctx context.Context, params GetPreviousCloseParams, opts ...client.Option) (*AggsResponse, error) {
```
morningvera, line 29: This is definitely up for debate. We ended up not following our current convention here (path params as inputs and query params as a struct ptr) for a couple reasons:
- some of our endpoints have a lot of path params; long method signatures seem harder to use and document
- using a type like this allows us to avoid calling `Sprintf` on our endpoint paths and we can set path and query params with resty instead

rest/polygon.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+// todo: add comments for godoc
+
+type polygonClient struct {
+	Aggregates *aggregates.Client
+	// todo: Trades, Quotes, etc
+}
+
+func New(config client.HTTPBaseConfig) *polygonClient {
```
qrpike, line 15: I don't want to have to specify the domain or max retries unless they are different than the default. Most of the time the domain and max retries are fine with default values. API key is the only part which will surely be required to be known to create a client. Can we lower the required information to create the client to only what is required. With alternative methods for overriding defaults.

rest/polygon_test.go
```diff
@@ -0,0 +1,121 @@
...
+	c := polygon.New(client.HTTPBaseConfig{
+		URL:        "https://api.polygon.io",
+		Key:        os.Getenv("API_KEY"),
+		MaxRetries: 3,
+	})
+
+	res, err := c.Aggregates.Get(context.Background(), aggregates.GetParams{
+		Ticker:     "AAPL",
+		Multiplier: 1,
+		Resolution: "day",
```
qrpike, line 28: I think Resolution should be a type. There are a fixed size of options for this, and sometimes its easy to forget if its "day" or "days". Having it be a type allows the IDE to list what is available while you're using the library which is nice DX. I think we should do this for most enum values. ( Sort direction, etc ).

rest/polygon_test.go
```diff
@@ -0,0 +1,121 @@
...
+func TestAggsGroupedDaily(t *testing.T) {
+	c := polygon.New(client.HTTPBaseConfig{
+		URL:        "https://api.polygon.io",
+		Key:        os.Getenv("API_KEY"),
+		MaxRetries: 3,
+	})
+
+	res, err := c.Aggregates.GetGroupedDaily(context.Background(), aggregates.GetGroupedDailyParams{
+		Locale:     "global",
+		MarketType: "crypto",
```
qrpike, line 81: These are more instances where it would be nice to have types. Matthew has generated a library with these, which would be nice to incorporate somehow ( potentially copy/open-source/something ) if possible.

go.mod
```diff
@@ -1,3 +1,29 @@
...
 
-go 1.13
+go 1.17
+
+require (
+	github.com/gin-gonic/gin v1.7.7
```
matichenor, line 6: Did we review the licenses on each of this libraries?  Do we have defined good/bad license types?

rest/aggregates/aggregates.go
```diff
@@ -0,0 +1,45 @@
+package aggregates
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/polygon-io/client-golang/rest/client"
+)
+
+// todo: add comments for godoc
```
matichenor, line 10: I can haz godoc?

morningvera: you sure can

go.mod
morningvera: yeah, they all have permissive licenses (MIT or BSD). not sure we need to worry about this just yet but there are tools out there to check dependency licenses

rest/polygon_test.go
morningvera: totally agree on this, we can run through and add types for any enum style params

rest/aggregates/aggregates.go
mmoghaddam385: Oops I think I was looking at an older commit. I like what you've done with the query param struct within the params struct

rest/aggs/aggs.go
```diff
@@ -0,0 +1,42 @@
...
+// todo: add comments for godoc
+
+type Client struct {
+	client.BaseClient
+}
+
+func (ac *Client) Get(ctx context.Context, params GetParams, opts ...client.Option) (*AggsResponse, error) {
+	res := &AggsResponse{}
+	path := "/v2/aggs/ticker/{ticker}/range/{multiplier}/{resolution}/{from}/{to}"
+	err := ac.Call(http.MethodGet, path, params, res, append([]client.Option{client.WithContext(ctx)}, opts...)...)
```
calebcase, line 19: Instead of having to do this context to options dance in a bunch of places should the Call method instead take a context as the first argument?

rest/aggs/models.go
```diff
@@ -0,0 +1,216 @@
...
+	return nil
+}
+
+type GetParams struct {
+	Ticker      string
+	Multiplier  int
+	Resolution  string
+	From        time.Time
+	To          time.Time
+	QueryParams *GetQueryParams
```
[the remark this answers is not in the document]
mcdayoub: Not every endpoint has path params, but the ones that do have them require them.

Constructing the params as-is looks like this:
```
	res, err := c.Aggs.GetGroupedDaily(context.Background(), aggs.GetGroupedDailyParams{
		Locale:     "global",
		MarketType: aggs.Crypto,
		Date:       time.Date(2021, 7, 21, 0, 0, 0, 0, time.Local),
		QueryParams: &aggs.GetGroupedDailyQueryParams{
			Adjusted: true,
		},
	})
```

If `Get{API}Params` had a struct for the path params, constructing that same request would look like this:

```
	res, err := c.Aggs.GetGroupedDaily(context.Background(), aggs.GetGroupedDailyParams{
		PathParams: {
                 	Locale:     "global",
			MarketType: aggs.Crypto,
			Date:       time.Date(2021, 7, 21, 0, 0, 0, 0, time.Local),
		},
		QueryParams: &aggs.GetGroupedDailyQueryParams{
			Adjusted: true,
		},
	})
```

It's pretty redundant to put the path params in the PathParams struct since they are already required when creating any request to this endpoint. None of the fields inside of PathParams can be left out.

For these reasons we decided to simplify constructing each query without `PathParams`.

This pattern works for each endpoint, so our docs can state 
```
All path params are required and are listed as `Get{API}Params` fields except for the optional `QueryParams` 
```

rest/aggs/aggs.go
morningvera: makes sense! just updated it

rest/client/client.go
morningvera: yeah, this is just the default. for now, i'm going to hardcode some of these defaults and we can figure out which ones to add to our request options later

rest/aggs/aggs.go
```diff
@@ -0,0 +1,42 @@
...
+
+// todo: add comments for godoc
+
+type Client struct {
+	client.BaseClient
+}
+
+func (ac *Client) Get(ctx context.Context, params GetParams, opts ...client.Option) (*AggsResponse, error) {
+	res := &AggsResponse{}
+	path := "/v2/aggs/ticker/{ticker}/range/{multiplier}/{resolution}/{from}/{to}"
```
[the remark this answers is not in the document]
morningvera: done

rest/aggs/models.go
```diff
@@ -0,0 +1,243 @@
...
+type AggsResponse struct {
+	client.BaseResponse
+	Ticker       string      `json:"ticker,omitempty"`
+	QueryCount   int         `json:"queryCount"`
+	ResultsCount int         `json:"resultsCount"`
+	Adjusted     bool        `json:"adjusted"`
+	Aggs         []Aggregate `json:"results,omitempty"`
+}
+
+func (r *AggsResponse) UnmarshalJSON(data []byte) error {
```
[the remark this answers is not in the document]
morningvera: yeah, implementing UnmarshalJSON is required if we want resty to automatically unmarshal the response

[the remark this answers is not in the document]
calebcase: I wasn't able to reproduce that behavior. Here's how I tried:

```
package main

import (
	"os"
	"time"

	"github.com/davecgh/go-spew/spew"
	"github.com/go-resty/resty/v2"
)

type Aggregate struct {
	Ticker            string  `json:"T,omitempty"`
	Volume            float64 `json:"v"`
	VWAP              float64 `json:"vw,omitempty"`
	AggregateVWAP     float64 `json:"a,omitempty"`
	Open              float64 `json:"o"`
	Close             float64 `json:"c"`
	High              float64 `json:"h"`
	Low               float64 `json:"l"`
	Timestamp         int64   `json:"t"`
	Transactions      int64   `json:"n,omitempty"`
	Market            string  `json:"m,omitempty"`
	Exchange          int32   `json:"x,omitempty"`
	Locale            string  `json:"g,omitempty"`
	OfficialOpenPrice float64 `json:"op,omitempty"`
	AverageSize       float64 `json:"z,omitempty"`
	AccumulatedVolume float64 `json:"av,omitempty"`
	StartTimestamp    int64   `json:"s,omitempty"`
	EndTimestamp      int64   `json:"e,omitempty"`
}

type AggsResponse struct {
	Ticker       string      `json:"ticker,omitempty"`
	QueryCount   int         `json:"queryCount"`
	ResultsCount int         `json:"resultsCount"`
	Adjusted     bool        `json:"adjusted"`
	Aggs         []Aggregate `json:"results,omitempty"`
}

func main() {
	apiKey := os.Getenv("POLYGON_API_KEY")

	rc := resty.New()
	rc.SetBaseURL("https://api.polygon.io")
	rc.SetAuthToken(apiKey)
	rc.SetRetryCount(3)
	rc.SetTimeout(10 * time.Second)

	ar := &AggsResponse{}
	req := rc.R().SetResult(ar)

	res, err := req.Execute("GET", "/v2/aggs/ticker/AAPL/prev")
	if err != nil {
		panic(err)
	}

	spew.Dump(res)
	spew.Dump(ar)
}
```

```
➜ ./resty 
(*resty.Response)(0xc00028a0a0)({"ticker":"AAPL","queryCount":1,"resultsCount":1,"adjusted":true,"results":[{"T":"AAPL","v":9.6917302e+07,"vw":156.1598,"o":158.93,"c":154.73,"h":159.28,"l":154.5,"t":1647032400000,"n":880727}],"status":"OK","request_id":"093d548a9a4ed255d27b049e7fb92222","count":1})
(*main.AggsResponse)(0xc00002a3c0)({
 Ticker: (string) (len=4) "AAPL",
 QueryCount: (int) 1,
 ResultsCount: (int) 1,
 Adjusted: (bool) true,
 Aggs: ([]main.Aggregate) (len=1 cap=4) {
  (main.Aggregate) {
   Ticker: (string) (len=4) "AAPL",
   Volume: (float64) 9.6917302e+07,
   VWAP: (float64) 156.1598,
   AggregateVWAP: (float64) 0,
   Open: (float64) 158.93,
   Close: (float64) 154.73,
   High: (float64) 159.28,
   Low: (float64) 154.5,
   Timestamp: (int64) 1647032400000,
   Transactions: (int64) 880727,
   Market: (string) "",
   Exchange: (int32) 0,
   Locale: (string) "",
   OfficialOpenPrice: (float64) 0,
   AverageSize: (float64) 0,
   AccumulatedVolume: (float64) 0,
   StartTimestamp: (int64) 0,
   EndTimestamp: (int64) 0
  }
 }
})
```

rest/aggs/aggs.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+// Client defines a REST client for the Polygon aggregates API.
+type Client struct {
+	client.BaseClient
+}
+
+// Get retrieves aggregate bars for a specified ticker over a given date range in custom time window sizes.
+// For example, if timespan = ‘minute’ and multiplier = ‘5’ then 5-minute bars will be returned.
+func (ac *Client) Get(ctx context.Context, params GetParams, opts ...client.Option) (*AggsResponse, error) {
```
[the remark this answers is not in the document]
calebcase: Where req has a resty request embedded.

rest/client/client.go
morningvera: yeah, quite possibly. i'm removing that option for now (advanced users can still set the request ID header if they now how). we can figure out if we want to expose it later

rest/aggs/models.go
```diff
@@ -0,0 +1,243 @@
...
+type GetParams struct {
+	Ticker      string
+	Multiplier  int
+	Resolution  Resolution
+	From        time.Time
+	To          time.Time
+	QueryParams *GetQueryParams
+}
+
+type GetQueryParams struct {
```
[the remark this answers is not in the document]
morningvera: done

rest/client/client.go
jbonzo: We block that in our ingress so even if they try it won't enter our system.

go.mod
jbonzo: why are we importing gin?

morningvera: we're not, this is an outdated comment

rest/client/client.go
morningvera: makes sense, we ended up removing this from the request options regardless

PR #11, opened 2022-03-16
started on aggs tests

mcdayoub:


rest/aggs_test.go
```diff
@@ -0,0 +1,209 @@
+package polygon_test
```
morningvera, line 1: i think we should move this file to the aggs dir and change this to `aggs_test`

rest/aggs_test.go
```diff
@@ -0,0 +1,209 @@
...
+		Ticker: "AAPL",
+		Date:   time.Date(2020, 10, 14, 0, 0, 0, 0, time.Local),
+		QueryParams: aggs.GetDailyOpenCloseQueryParams{
+			Adjusted: &adjusted,
+		},
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	assert.Equal(t, "AAPL", res.Symbol)
```
morningvera, line 208: not sure this is necessary but we could compare the entire response against an expected result instead of just the ticker

rest/aggs_test.go
```diff
@@ -0,0 +1,209 @@
...
+	res, err := c.Aggs.Get(context.Background(), aggs.GetParams{
+		Ticker:     "AAPL",
+		Multiplier: 1,
+		Resolution: "day",
+		From:       time.Date(2021, 7, 22, 0, 0, 0, 0, time.Local),
+		To:         time.Date(2021, 8, 22, 0, 0, 0, 0, time.Local),
+		QueryParams: aggs.GetQueryParams{
+			Adjusted: polygon.Bool(true),
+			Sort:     polygon.AggsSort(aggs.Desc),
+			Limit:    polygon.Int32(1),
```
morningvera, line 60: we could add the explain param here if we wanna get that sweet 100% coverage 😎

morningvera:
don't forget to uncomment this: https://github.com/polygon-io/client-go/blob/v1/.github/workflows/test.yml#L23-L24

morningvera (approved):

PR #12, opened 2022-03-17
add mit license

morningvera:
i'm assuming we want an MIT license here (copied from go-app-ticker-wall)


clickingbuttons (approved):
I agree this is the best license for this

PR #13, opened 2022-03-21
add list quotes with an iterator pattern

morningvera:
This PR implements an iterator pattern that we'll reuse for all the other list endpoints. The `Iter` type is designed to be reusable so definitely looking for a good review of that part.


rest/client/client.go
```diff
@@ -87,6 +97,17 @@ type BaseResponse struct {
 	PaginationHooks
 }
 
+// PaginationHooks are links to next and/or previous pages. Embed this struct into your API response if
+// your endpoint is going to support pagination.
```
mcdayoub, line 101: I'm wondering about the phrasing here
Should `your endpoint` be `the endpoint` ? If I was a user to this I'm not sure if the user would consider possession of the endpoint

rest/client/client.go
```diff
@@ -16,10 +17,25 @@ const (
...
 )
 
-// Params defines an interface that path parameter and query parameter types must implement.
+// Params defines an interface that parameter types must implement.
 type Params interface {
+	// Path returns a map of URL path parameters.
 	Path() map[string]string
-	Query() map[string]string
+	// Query returns query string parameters as URL values.
+	Query() url.Values
```
morningvera, line 25: ended up changing this to make it easier to use for both get and list methods

rest/client/iter.go
```diff
@@ -0,0 +1,54 @@
...
+	if len(it.results) == 0 && it.page.NextPageURL() != "" {
+		it.page, it.results, it.err = it.query(it.page.NextPageURL())
+	}
+	if it.err != nil || len(it.results) == 0 {
+		return false
+	}
+	it.current = it.results[0]
+	it.results = it.results[1:]
+	return true
+}
```
jbonzo, line 39: These aren't the same semantics we use for our own internal clients so I would like more edge testing to gain confidence.

jbonzo (commented):
How will you you implement iteration for our older APIs that don't follow the conventional pagination techniques. Such as aggs

rest/client/iter.go
morningvera: makes sense! i based this partially on our own pattern and partially on [stripe's](https://github.com/stripe/stripe-go/blob/master/iter.go#L62). i was planning to add some tests for this iter type itself so i'll go ahead and do that in this PR

morningvera: just added some tests for the iterator pattern

rest/client/iter.go
```diff
@@ -0,0 +1,54 @@
...
+
+// Page returns the current results page in the form of an API response.
+func (it *Iter) Page() ListResponse {
+	return it.page
+}
+
+// Current returns the result that the iterator is currently pointing to.
+func (it *Iter) Current() interface{} {
```
calebcase, line 47: I'm not sure there's a consistent naming preference out there. I've seen `Item()` or `Value()`. I think those might be clearer than `Current()`.

FWIW, this Iter and the desire to make a "common" one is probably ripe for generics use. We side step this internally by generating code that is type specific. I think we should either have generated type specific code here or use generics.

rest/client/iter.go
morningvera: @calebcase makes sense, i'll add that

rest/client/iter.go
jbonzo, line 47: I don't love this returning an empty interface vs it returning the explicit type. I get there is a desire to have a "common" iter but I rather not have to type assert everywhere

morningvera: see [here](https://github.com/polygon-io/client-go/pull/13/files#diff-c2e88127eb60436aa954e3dd31a34149c0103930e39213858a45c48ce3c04a0eR17). this method is only meant for use by domain specific iters and users should call `iter.Quote()` instead

jbonzo: Is there danger of that assertion failing? If not then this is fine

morningvera: stripe uses `Current()` but i think i'll change it to `Value()`. also, i generally agree that generics could make sense here but i don't think we're trying to use that feature heavily just yet

morningvera: if someone were to call `iter.Current()` and do the type assertion themselves (instead of `iter.Quote()` which handles nil) before `iter.Next()`, it'll return nil which would make the assertion fail. seems like it would be a deliberate choice by a user though. it's done that way so that people don't have to handle the first item in the list before starting the loop

calebcase: Now that I've seen how it's used in Quotes I think my concerns about it being held wrong are resolved.

calebcase (approved):

PR #15, opened 2022-03-23
move models to a separate package

morningvera:
Moving our models to a separate package. This makes more sense for eventual code generation and lets users just import two packages to use our client (`polygon` and `models`).

### DIFF
--- a/rest/models/types.go
+++ b/rest/models/types.go
@@ -39,7 +39,17 @@ const (
 	Desc Order = "desc"
 )
 
-// Ptr returns a pointer to a specified value.
-func Ptr[T any](v T) *T {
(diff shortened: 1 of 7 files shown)


rest/models/types.go
```diff
@@ -0,0 +1,45 @@
...
+// desc will return results in descending order (newest at the top).
+type Order string
+
+const (
+	Asc  Order = "asc"
+	Desc Order = "desc"
+)
+
+// Ptr returns a pointer to a specified value.
+func Ptr[T any](v T) *T {
```
morningvera, line 43: pls can we use generics?? 🙂

mcdayoub (approved):

PR #19, opened 2022-03-28
refactor ptr methods to something generic

morningvera:
the v3 golangci-lint gh action works with basic generics!

### DIFF
--- a/rest/models/types.go
+++ b/rest/models/types.go
@@ -46,29 +46,7 @@ const (
 	Desc Order = "desc"
 )
 
-// todo: replace these with a single generic method once the linter supports it
-
-// Bool returns a pointer to a bool value.
-func Bool(v bool) *bool {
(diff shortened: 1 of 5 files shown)


rest/models/types.go
```diff
@@ -48,27 +48,7 @@ const (
...
-
-// OrderBy returns a pointer to an order value.
-func OrderBy(v Order) *Order {
-	return &v
-}
-
-// SortOn returns a pointer to a sort value.
-func SortOn(v Sort) *Sort {
+// Any returns a pointer to any value.
+func Any[T any](v T) *T {
```
morningvera, line 52: lmk if something like `Ptr` is a better name for this

morningvera: never mind, i already changed it to ptr

mcdayoub (approved):

PR #20, opened 2022-03-28
change query params to proper timestamps

morningvera:


rest/client/client.go
```diff
@@ -104,8 +104,7 @@ type BaseResponse struct {
 // PaginationHooks are links to next and/or previous pages. Embed this struct into an API response if
 // the endpoint supports pagination.
 type PaginationHooks struct {
-	NextURL     string `json:"next_url,omitempty"`
-	PreviousURL string `json:"previous_url,omitempty"`
```
morningvera, line 108: removing this since our api doesn't actually support it (we could always add it back later)

mcdayoub: ah darn I thought it was cool but best to not confuse anyone

PR #24, opened 2022-03-30
V.reftickers

morningvera:


rest/models/tickers.go
```diff
@@ -0,0 +1,298 @@
...
+
+// ListTickersQueryParams is the set of query parameters for requesting reference tickers.
+type ListTickersQueryParams struct {
+	TickerEQ  *string
+	TickerLT  *string
+	TickerLTE *string
+	TickerGT  *string
+	TickerGTE *string
+
+	// todo: which ones should be enums?
```
morningvera, line 91: i'm running into more query param problems as we add more endpoints:
- some endpoints expect `fx` instead of `forex` for market type
- asset class is very similar to market type but also slightly different
- ticker types lets you query on asset class but only `stocks` returns results
- some endpoints have after primary and i don't know what that is

mcdayoub: at least you know what `after primary` is now

mcdayoub (approved):

PR #25, opened 2022-03-30
add automatic validation and query encoding

morningvera:
Sorry in advance for the large PR. This is a pretty big shift and there are a couple things that definitely aren't perfect:
- Time/date encoding is a problem. Most of the newer endpoints encode to nanoseconds, but reference tickers encodes to "2006-01-02". For now, I switched those reference tickers params to strings. Not a permanent solution by any means.
- Had to change up the iterator instantiation a bit. The additional error handling feels a bit messy. At least we no longer have to implement the `String()` method though.
- Probably something else I'm missing.

Otherwise, I feel like this significantly reduces our maintenance burden since types only need to implement a `Path()` method now. It also feels a bit easier to use since we removed the need to nest query param structs when making calls.


rest/models/aggs.go
```diff
@@ -49,25 +47,21 @@ type AggsResponse struct {
...
+// GetAggsParams is the set of path and query parameters for requesting aggs.
 type GetAggsParams struct {
-	Ticker      string
-	Multiplier  int
-	Resolution  Resolution
-	From        time.Time
-	To          time.Time
-	QueryParams GetAggsQueryParams
-}
+	Ticker     string     `validate:"required"`
```
mcdayoub, line 52: I already like this more than having an embedded `QueryParams` struct. 

The user should only have to care about what's required, and not thinking:
 "if it's in the path then it's required, if it's in the `QueryParams` it's not required. Also the path is everything in `GetAggsParams` except for the embedded `QueryParams`. Seeing `required` explicitly listed here saves a lot of thinking from the user.

rest/models/quotes.go
```diff
@@ -42,100 +38,25 @@ type QuotesResponse struct {
...
-	TimestampGTE *time.Time
-
-	Order *Order
-
-	Limit *int
-
-	Sort *Sort
+	Ticker string `validate:"required"`
 
-	AfterPrimary   *string
```
mcdayoub, line 63: I'd like to know if there's any reason why we should keep `AfterPrimary` `AfterSecondary` and `Cursor`

mcdayoub (commented):
This is a huge cleanup. I like the new way to construct the `Params` for each function. It's similar to how people interact with our docs too.

rest/models/quotes.go
morningvera: yeah, i'm not sure but it's something we can discuss this afternoon

mcdayoub (approved):

PR #26, opened 2022-03-31
improve iterator methods

morningvera:
made some improvements with the generic iter branch that I want to fold into v1 since we're not using generics yet

### DIFF
--- a/rest/models/quotes.go
+++ b/rest/models/quotes.go
@@ -33,7 +33,7 @@ type Quote struct {
 // QuotesResponse is returned by the list quotes API and contains a list of quotes for the specified ticker.
 type QuotesResponse struct {
 	client.BaseResponse
-	Results []*Quote `json:"results,omitempty"`
+	Results []Quote `json:"results,omitempty"`
 }
 
 // ListQuotesParams is the set of path and query parameters that are used when requesting quotes via the ListQuotes method.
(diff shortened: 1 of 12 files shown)


rest/models/quotes.go
```diff
@@ -33,7 +33,7 @@ type Quote struct {
 // QuotesResponse is returned by the list quotes API and contains a list of quotes for the specified ticker.
 type QuotesResponse struct {
 	client.BaseResponse
-	Results []*Quote `json:"results,omitempty"`
+	Results []Quote `json:"results,omitempty"`
```
morningvera, line 36: most of the changes involve refactoring these `[]*T` fields (i don't like them)

rest/quotes/quotes.go
```diff
@@ -39,30 +35,30 @@ func (it *QuotesIter) Quote() *models.Quote {
...
-		err := c.Call(ctx, http.MethodGet, url, nil, res, options...)
-
-		results := make([]interface{}, len(res.Results))
-		for i, v := range res.Results {
-			results[i] = v
-		}
-
-		return res, results, err
-	})
+	url, err := c.EncodeParams(models.ListQuotesPath, params)
```
morningvera, line 45: i used a method like this in the generic iter branch and i think it's a bit cleaner than the way i did things before (i didn't like that NewIter was a method implemented on the client type)

mcdayoub (approved):

PR #27, opened 2022-03-31
add path param encoding

morningvera:
this includes some light standardization of type names and godoc comments (preparing for code gen). luckily, all of the time path params work with the "2006-01-02" format (at least for what is currently implemented). going to try to improve our handling of time in a future PR


rest/aggs/aggs_test.go
```diff
@@ -46,7 +46,7 @@ func TestGetAggs(t *testing.T) {
 	httpmock.ActivateNonDefault(c.Aggs.HTTP.GetClient())
 	defer httpmock.DeactivateAndReset()
 
-	httpmock.RegisterResponder("GET", "https://api.polygon.io/v2/aggs/ticker/AAPL/range/1/day/1626912000000/1629590400000?adjusted=true&explain=false&limit=1&sort=desc",
+	httpmock.RegisterResponder("GET", "https://api.polygon.io/v2/aggs/ticker/AAPL/range/1/day/2021-07-22/2021-08-22?adjusted=true&explain=false&limit=1&sort=desc",
```
morningvera, line 49: not sure if we should just leave this as is or change it back to millis once we figure out the time problem

rest/client/client.go
```diff
@@ -18,12 +18,6 @@ const (
 	DefaultRetryCount = 3
 )
 
-// Params defines an interface that parameter types must implement.
-type Params interface {
```
morningvera, line 22: 👋

rest/client/client.go
```diff
@@ -129,6 +138,55 @@ func (c *Client) newRequest(ctx context.Context, params Params, response interfa
...
+	}
+
+	pathParams := map[string]string{}
+	for k, v := range val {
+		pathParams[k] = v[0] // only accept the first one for a given key
+	}
+	return pathParams, nil
+}
+
+func (c *Client) encodePathString(uri string, params interface{}) (string, error) {
```
morningvera, line 161: this is all getting a bit messy since `newRequest` and `EncodeParams` do things differently. kinda wondering if we should just stop using resty to encode path and query params and do it ourselves since we already have to for iter methods

mcdayoub (approved):

PR #30, opened 2022-04-01
snapshot all tickers query param

mcdayoub:


rest/models/snapshot.go
```diff
@@ -12,6 +12,8 @@ const (
 type GetAllTickersSnapshotParams struct {
 	Locale     MarketLocale `validate:"required" path:"locale"`
 	MarketType MarketType   `validate:"required" path:"marketType"`
+
+	Tickers *string `query:"tickers"`
```
morningvera, line 16: should it be `*[]string`?

morningvera (approved):

PR #36, opened 2022-04-06
Darcy.market status

Darcy-Linde:


rest/models/markets.go
```diff
@@ -11,6 +12,13 @@ type GetMarketHolidaysParams struct{}
 // GetMarketHolidaysResponse is the response returned by the GetMarketHolidays method.
 type GetMarketHolidaysResponse []MarketHoliday
 
+// todo: this endpoint is unlikely to ever have params so should we delete this type?
+// GetMarketStatusParams is the set of parameters for the GetMarketStatus method.
+type GetMarketStatusParams struct{}
+
+// GetMarketStatusResponse is the response returned by the GetMarketStatus method.
+type GetMarketStatusResponse MarketStatus
```
morningvera, line 20: i think we should delete the `MarketStatus` type and move its fields into this struct instead (`MarketHoliday` was only done this way because the response is a slice of market holidays)

rest/models/markets.go
```diff
@@ -11,6 +12,13 @@ type GetMarketHolidaysParams struct{}
 // GetMarketHolidaysResponse is the response returned by the GetMarketHolidays method.
 type GetMarketHolidaysResponse []MarketHoliday
 
+// todo: this endpoint is unlikely to ever have params so should we delete this type?
+// GetMarketStatusParams is the set of parameters for the GetMarketStatus method.
+type GetMarketStatusParams struct{}
```
morningvera, line 17: i'm thinking we should delete these after a little more thought. wanna remove this and the market holiday params?

Darcy-Linde: Sure I'll remove them

morningvera (approved):

PR #40, opened 2022-04-07
Expose NewRequest for Builder pattern

clickingbuttons:
This allows building custom requests using Resty's builder pattern. This is most useful for query params and can replace the varargs `...opts` for setting HTTP headers.

We already expose `resty.Client` in `rest/client/client.go` so I figured why not expose NewRequest (which is a `*resty.Request`) and tweak it to allow `params` to be a `map[string]string`.


rest/client/client_test.go
```diff
@@ -0,0 +1,9 @@
+package client_test
+
+import (
+	"testing"
+)
+
+func TestGetAggs(t *testing.T) {
+
```
morningvera, line 8: guessing this can be deleted since you added a test in aggs

rest/polygon.go
```diff
@@ -15,16 +15,18 @@ type polygonClient struct {
 	Reference *reference.Client
 	Trades    *trades.Client
 	Snapshot  *snapshot.Client
+	Client    *client.Client
```
morningvera, line 18: you could "inherit" here (`client.Client` instead of `Client *client.Client`). that would make using it a bit cleaner (`c.NewRequest(...)` instead of `c.Client.NewRequest(...)`)

rest/client/client.go
```diff
@@ -100,11 +101,16 @@ func (c *Client) EncodeParams(uri string, params interface{}) (string, error) {
...
+//nolint:nestif // 2 levels is okay <3
+func (c *Client) NewRequest(ctx context.Context, params, response interface{}) (*resty.Request, error) {
 	req := c.HTTP.R().SetContext(ctx)
-	if params != nil {
+	req.SetResult(response).SetError(&models.ErrorResponse{})
+	if params == nil {
+		return req, nil
+	}
+
+	v := reflect.ValueOf(params)
```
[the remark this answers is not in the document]
clickingbuttons: I was trying a solution that didn't require a struct but I'll revert because it is rather ugly.

rest/aggs/aggs_test.go
```diff
@@ -66,7 +66,39 @@ func TestGetAggs(t *testing.T) {
...
+		Limit:      models.Ptr(1),
+		Explain:    models.Ptr(false),
+	}, res)
+	// Allows using uncodumented params
+	req.SetQueryParam("customparam", "customvalue")
+	// ...or setting arbitrary headers
+	req.SetHeader("customheader", "customvalue")
+	assert.Nil(t, err)
+
+	_, err = req.Execute(http.MethodGet, models.GetAggsPath)
```
morningvera, line 97: it might be nice to build a request that already has these two inputs handled for the user so they could just call `req.Send()` once they're done adding custom stuff

morningvera: we discussed irl and this is probably not worth the extra work

rest/trades/trades_test.go
```diff
@@ -165,7 +165,7 @@ func TestGetLastCryptoTrade(t *testing.T) {
 	assert.Equal(t, expectedResponse, string(b))
 }
 
-func registerResponder(url string, body string) {
+func registerResponder(url, body string) {
```
mcdayoub, line 168: Yeah open another PR for anything that's specifically cleanup related so we can merge that no question

clickingbuttons:
Closing in favor of a method that support `List` endpoints

PR #45, opened 2022-04-07
Darcy.exchanges

Darcy-Linde:


rest/reference/exchanges.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"context"
+	"net/http"
+
+	"github.com/polygon-io/client-go/rest/client"
+	"github.com/polygon-io/client-go/rest/models"
+)
+
+// ListExchangesIter is an iterator for the ListExchanges method.
+type ListExchangesIter struct {
```
morningvera, line 12: i'm pretty sure this isn't a list/iterator method (the response would include a next URL if it was)

Darcy-Linde: Ah I got tunnel visioned because it was in a results array. I'll fix it.

rest/reference/exchanges_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestListExchanges(t *testing.T) {
+	c := polygon.New("API_KEY")
+
+	httpmock.ActivateNonDefault(c.Reference.HTTP.GetClient())
+	defer httpmock.DeactivateAndReset()
+
+	exchange1 := `{
```
morningvera, line 20: might be better to just put this directly in `expectedResponse` since you don't need to verify it by itself anymore

morningvera (approved):
one minor comment but otherwise lgtm

PR #47, opened 2022-04-07
V.time

mcdayoub:


rest/models/aggs.go
```diff
@@ -16,8 +12,8 @@ type GetAggsParams struct {
 	Ticker     string     `validate:"required" path:"ticker"`
 	Multiplier int        `validate:"required" path:"multiplier"`
 	Resolution Resolution `validate:"required" path:"resolution"`
-	From       time.Time  `validate:"required" path:"from"`
-	To         time.Time  `validate:"required" path:"to"`
+	From       *Millis    `validate:"required" path:"from"`
```
morningvera, line 15: i think we can actually avoid this pointer by changing our input to `RegisterCustomTypeFunc`. thinking we could check `time.IsZero()` and return "" if so

mcdayoub: Talked on slack, this didn't work because it produces some weird output in the tests
```
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -16,3 +16,5 @@
        	            	 			"t": 1577941200000,
        	            	-			"n": 1
        	            	+			"n": 1,
        	            	+			"s": -62135596800000,
        	            	+			"e": -62135596800000
```

rest/models/types.go
```diff
@@ -73,3 +79,57 @@ const (
 	Gainers Direction = "gainers"
 	Losers  Direction = "losers"
 )
+
+// todo: godoc
```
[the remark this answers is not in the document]
mcdayoub: I'll merge this PR and then do a separate PR with the godoc stuff

PR #48, opened 2022-04-08
add query building for trades and quotes

morningvera:


rest/quotes/quotes.go
```diff
@@ -41,7 +41,7 @@ func (it *ListQuotesIter) Quote() models.Quote {
 //   if iter.Err() != nil {
 //       return err
 //   }
-func (c *Client) ListQuotes(ctx context.Context, params models.ListQuotesParams, options ...models.RequestOption) (*ListQuotesIter, error) {
+func (c *Client) ListQuotes(ctx context.Context, params *models.ListQuotesParams, options ...models.RequestOption) (*ListQuotesIter, error) {
```
morningvera, line 44: if we want to add this builder pattern, we should probably do it sooner than later since it kinda requires this interface change

rest/models/quotes.go
```diff
@@ -19,9 +19,40 @@ type ListQuotesParams struct {
...
-	Order        *Order     `query:"order"`
-	Limit        *int       `query:"limit"`
-	Sort         *Sort      `query:"sort"`
+
+	Sort  *Sort  `query:"sort"`
+	Order *Order `query:"order"`
+	Limit *int   `query:"limit"`
+}
+
+func (p ListQuotesParams) WithTimestamp(c Comparator, q time.Time) *ListQuotesParams {
```
morningvera, line 28: intentionally choosing not to document these since the model itself should handle it

mcdayoub (approved):

PR #50, opened 2022-04-08
refactor public interface of client

morningvera:
trying to reduce the chances for varying behavior between our normal get and list methods. also wanted to remove the newRequest helper so people aren't tempted to flip it to public 😄 also just changing url vars to uri (technically more correct i think) so we never have to worry about collisions with the `net/url` package.

### DIFF
--- a/rest/models/request.go
+++ b/rest/models/request.go
@@ -20,15 +20,15 @@ type RequestOptions struct {
 // RequestOption changes the configuration of RequestOptions.
 type RequestOption func(o *RequestOptions)
 
-// WithAPIKey sets the APIKey for an Option.
-func WithAPIKey(id string) RequestOption {
+// APIKey sets an APIKey as an option.
+func APIKey(id string) RequestOption {
 	return func(o *RequestOptions) {
 		o.APIKey = &id
 	}
 }
 
-// WithHeader sets a Header for an Option.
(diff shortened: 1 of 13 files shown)


rest/models/request.go
```diff
@@ -20,15 +20,15 @@ type RequestOptions struct {
 // RequestOption changes the configuration of RequestOptions.
 type RequestOption func(o *RequestOptions)
 
-// WithAPIKey sets the APIKey for an Option.
-func WithAPIKey(id string) RequestOption {
+// APIKey sets an APIKey as an ption.
+func APIKey(id string) RequestOption {
```
morningvera, line 24: this is sort of a builder pattern but it doesn't appear that way to users so With felt somewhat confusing

rest/models/request.go
```diff
@@ -20,15 +20,15 @@ type RequestOptions struct {
 // RequestOption changes the configuration of RequestOptions.
 type RequestOption func(o *RequestOptions)
 
-// WithAPIKey sets the APIKey for an Option.
-func WithAPIKey(id string) RequestOption {
+// APIKey sets an APIKey as an ption.
```
mcdayoub, line 23: ption -> option

mcdayoub (approved):
just that one comment

PR #51, opened 2022-04-08
add query builder for aggs and snapshot

morningvera:


rest/models/aggs.go
```diff
@@ -15,12 +15,32 @@ type GetAggsParams struct {
 	From       Millis     `validate:"required" path:"from"`
 	To         Millis     `validate:"required" path:"to"`
 
-	Sort     *Order `query:"sort"`
```
morningvera, line 18: thoughts on this? i'm on the fence about it. on one hand, it will look different than our docs but on the other hand, people might be confused that they have to call `WithSort(q Order)`.

mcdayoub: In this PR I'm seeing the 

```
func (p GetAggsParams) WithOrder(q Order) *GetAggsParams {
	p.Order = &q
	return &p
}
```

So should the `WithOrder` func actually be called `WithSort`?

morningvera: if we follow the convention (field names == query tags), then it'll look like `WithSort(q Order)` because this endpoint uses the wrong query param name. should we just leave that confusion in there or break the convention and name this field `Order` (which is different from our docs)

mcdayoub (approved):

rest/models/aggs.go
morningvera: we decided to go with this change

PR #52, opened 2022-04-08
add query builders for reference

morningvera:


rest/models/types.go
```diff
@@ -6,11 +6,6 @@ import (
 	"time"
 )
 
-// Ptr returns a pointer to any value.
-func Ptr[T any](v T) *T {
```
morningvera, line 10: we can't have this is we're not mandating go1.18

rest/models/tickers.go
```diff
@@ -19,9 +19,9 @@ type ListTickersParams struct {
 	// todo: which ones should be enums?
 	Type     *string     `query:"type"`
 	Market   *MarketType `query:"market"` // todo: this endpoint apparently expects fx instead of forex
-	Exchange *string     `query:"exchange"`
-	CUSIP    *string     `query:"cusip"`
-	CIK      *string     `query:"cik"`
+	Exchange *int        `query:"exchange"`
```
morningvera, line 22: i think these should be integers (not sure why they were strings originally)

rest/models/tickers.go
```diff
@@ -16,13 +16,12 @@ type ListTickersParams struct {
...
 	TickerGTE *string `query:"ticker.gte"`
 
-	// todo: which ones should be enums?
 	Type     *string     `query:"type"`
-	Market   *MarketType `query:"market"` // todo: this endpoint apparently expects fx instead of forex
-	Exchange *string     `query:"exchange"`
-	CUSIP    *string     `query:"cusip"`
-	CIK      *string     `query:"cik"`
-	Date     *string     `query:"date"` // todo: this is "2006-01-02" format, need to figure out the best way to encode this without interfering with the default
+	Market   *AssetClass `query:"market"`
```
morningvera, line 20: using asset class here since that has "fx"

mcdayoub (approved):

PR #53, opened 2022-04-08
cleanups

mcdayoub:


rest/models/types.go
```diff
@@ -80,6 +80,55 @@ const (
...
+	DataTypeTrade DataType = "trade"
+	DataTypeBBO   DataType = "bbo"
+	DataTypeNBBO  DataType = "nbbo"
+)
+
+// SIP is the type of Securies Information Processor.
+type SIP string
+
+const (
+	CTA  DataType = "CTA"
```
morningvera, line 106: `DataType` -> `SIP`

mcdayoub: nice catch

rest/reference/market_test.go
```diff
@@ -20,33 +20,33 @@ func TestGetMarketHolidays(t *testing.T) {
 	{
 		"exchange": "NYSE",
 		"name": "Thanksgiving",
-		"date": "2020-11-26T00:00:00.000Z",
```
morningvera, line 23: hmm, wasn't this copied directly from our api response? i feel like we should conform to that instead of changing the test

rest/models/types.go
```diff
@@ -80,6 +80,55 @@ const (
 	Losers  Direction = "losers"
 )
 
+// AssetClass is an identifier for a group of similar financial instruments.
+type AssetClass string
+
+const (
+	AssetClassStocks  AssetClass = "stocks"
```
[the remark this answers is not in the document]
mcdayoub: In this case we already had a `Stocks` and `Crypto` const already. Any advice for handling that?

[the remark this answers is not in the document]
morningvera: ahh that's right, maybe `AssetStocks` would look a bit better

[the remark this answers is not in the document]
mcdayoub: perfect

rest/reference/market_test.go
morningvera: cool i'm fine with it as long as we unmarshal the real responses correctly

rest/models/types.go
```diff
@@ -80,6 +80,55 @@ const (
...
+	AssetOptions AssetClass = "options"
+	AssetCrypto  AssetClass = "crypto"
+	AssetFx      AssetClass = "fx"
+)
+
+// DataType is the type of data.
+type DataType string
+
+const (
+	DataTypeTrade DataType = "trade"
```
morningvera, line 97: can we delete `DataType` and `DividendType` from these consts?

rest/reference/market_test.go
mcdayoub: yeah I checked prod and currently every `date` in the response has the YYYY-MM-DD format

rest/models/types.go
mcdayoub: same issue as `AssetClass` where we already have a const for `Trade`

mcdayoub: For `DividendType` we already have a `LT` const too

morningvera: ahh, okay let's at least match the `AssetStocks` pattern then

morningvera (approved):

PR #55, opened 2022-04-08
add remaining time stuff

morningvera:


rest/client/client.go
```diff
@@ -40,16 +40,25 @@ func New(apiKey string) Client {
 	pe := form.NewEncoder()
 	pe.SetMode(form.ModeExplicit)
 	pe.SetTagName("path")
+	pe.RegisterCustomTypeFunc(func(x interface{}) ([]string, error) {
```
morningvera, line 43: might as well add these encoders in case we ever need them in the future

PR #61, opened 2022-04-11
reorganize iter and encoder types

morningvera:
i wanted to clean up the iter interface: `NewIter(ctx context.Context, path string, params interface{}, query Query) Iter`. to get there, it made sense to break the iter and encoder stuff into separate packages.

### DIFF
--- a/rest/encoder/encoder.go
+++ b/rest/encoder/encoder.go
@@ -1,0 +1,103 @@
+package encoder
+
+import (
+	"fmt"
+	"net/url"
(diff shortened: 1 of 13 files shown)


rest/encoder/encoder.go
```diff
@@ -0,0 +1,103 @@
+package encoder
```
mcdayoub, line 1: big fan of moving the encoder to its own package

mcdayoub (approved):

PR #66, opened 2022-04-12
add basic websocket client with second aggs

morningvera:
Still have a few todo comments to think about but I want to get a review on the general structure, public interface, and some of the logic. Going to finish adding godoc stuff after we get close to a good interface (but before merging).

### DIFF
--- a/websocket/polygon.go
+++ b/websocket/polygon.go
@@ -1,0 +1,63 @@
+package polygonws
+
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/polygon-io/client-go/websocket/client"
+	"github.com/polygon-io/client-go/websocket/models"
+)
+
+// PolygonClient defines a client to the Polygon WebSocket API.
+type PolygonClient struct {
+	client.Client
+}
+
+func New(apiKey string) *PolygonClient {
+	c := client.New(apiKey)
+	return &PolygonClient{
+		Client: c,
+	}
+}
+
+// StreamSecondAggs streams real-time second aggregates for a given ticker symbol. "*" will stream data for
+// all tickers of a given market type. For more details see https://polygon.io/docs/stocks/ws_stocks_a.
+// Callers of this method must close the connection:
+//   conn, aggs, err := c.StreamSecondAggs(context.TODO(), models.Stocks, "AAPL")
+//   if err != nil {
+//       return err
+//   }
+//   defer conn.Close()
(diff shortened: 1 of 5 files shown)


websocket/polygon.go
```diff
@@ -0,0 +1,63 @@
...
+
+// StreamSecondAggs streams real-time second aggregates for a given ticker symbol. "*" will stream data for
+// all tickers of a given market type. For more details see https://polygon.io/docs/stocks/ws_stocks_a.
+// Callers of this method must close the connection:
+//   conn, aggs, err := c.StreamSecondAggs(context.TODO(), models.Stocks, "AAPL")
+//   if err != nil {
+//       return err
+//   }
+//   defer conn.Close()
+func (pc *PolygonClient) StreamSecondAggs(ctx context.Context, market models.MarketType, ticker string) (*client.Conn, models.SecondAggs, error) {
```
morningvera, line 31: i think we need to create and return a new connection here because a connection should only handle one type of data and we need the user to call `defer conn.Close()` on it. this method handles creating a channel buffer for the user (i could also see arguments for letting the user do that and pass it in). it also assumes that users will stop the connection/loop with a context. definitely looking for feedback on this part!

websocket/polygon.go
```diff
@@ -0,0 +1,63 @@
...
+	conn, err := pc.Connect(market, "A."+ticker)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	data := make(chan []byte, 10000)
+	go conn.Collect(data)
+
+	secondAggs := make(models.SecondAggs, 10000)
+	// go conn.Process(ctx, data, secondAggs)
```
morningvera, line 41: this would be super nice but i can't think of a good way to do it at the moment. generics could work.

websocket/polygon.go
```diff
@@ -0,0 +1,63 @@
...
+
+	go func() {
```
morningvera, line 43: could use a sanity check on my logic here

websocket/polygon_test.go
```diff
@@ -0,0 +1,29 @@
...
+	"fmt"
+	"os"
+	"testing"
+	"time"
+
+	polygonws "github.com/polygon-io/client-go/websocket"
+	"github.com/polygon-io/client-go/websocket/models"
+)
+
+func TestMain(t *testing.T) {
```
morningvera, line 14: `API_KEY=YOUR_API_KEY go test -race -v .` if you wanna try running this. this is just temporary as we figure out the best way to unit test.

PR #68, opened 2022-04-15
add basic websocket client

morningvera:
For now, this client just connects and authenticates. Reconnects and subscription stuff will be added in future PRs. Looking for a review of the general structure, public interface and some of the internal logic (e.g. graceful disconnects).


websocket/polygon_test.go
```diff
@@ -0,0 +1,41 @@
...
+	"context"
+	"os"
+	"testing"
+	"time"
+
+	polygonws "github.com/polygon-io/client-go/websocket"
+	"github.com/sirupsen/logrus"
+)
+
+func TestMain(t *testing.T) {
```
morningvera, line 13: you can run this via `API_KEY=YOUR_API_KEY go test -v -race .`

websocket/polygon.go
```diff
@@ -0,0 +1,228 @@
...
+		c.log.Infof("authentication successful")
+	case "auth_failed":
+		c.log.Errorf("authentication failed, closing connection")
+		c.Close()
+		return nil
+	case "success":
+		c.log.Infof("subscription successful") // todo: can subscriptions fail?
+	default:
+		c.log.Debugf("unknown status message '%v'", cm.Status)
+	}
```
mmoghaddam385, line 225: Curious about the log levels here. I would put the "* successful" messages as debug level and the "unknown status message" as info level. My thought being success is more or less expected, while an unknown status message might be something you'd want to see in production

websocket/polygon.go
```diff
@@ -0,0 +1,228 @@
...
+	go c.process()
+
+	return c, nil
+}
+
+// todo: Subscribe, Unsubscribe, etc
+
+func (c *Client) Close() error {
+	c.cancel()
+	err := c.conn.WriteControl(websocket.CloseMessage, websocket.FormatCloseMessage(websocket.CloseNormalClosure, ""), time.Now().Add(writeWait))
```
mmoghaddam385, line 81: There's technically a race condition between this and the writer process. Although practically I'm not sure if it will really matter, is this something that you'd considered and decided to go with?

websocket/polygon.go
```diff
@@ -0,0 +1,228 @@
...
+			return
+		case <-ticker.C:
+			if err := c.conn.SetWriteDeadline(time.Now().Add(writeWait)); err != nil {
+				c.log.Errorf("failed to set write deadline: %v", err)
+				return // todo: should this return?
+			}
+			if err := c.conn.WriteMessage(websocket.PingMessage, nil); err != nil {
+				c.log.Errorf("failed to send ping message: %v", err)
+				return
+			}
```
mmoghaddam385, line 137: I think you can use `conn.WriteControl` with the `PingMessage` type here to simplify this into one function call

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/jarcoal/httpmock v1.1.0
 	github.com/sirupsen/logrus v1.8.1
 	github.com/stretchr/testify v1.7.0
+	go.uber.org/zap v1.21.0
```
mmoghaddam385, line 13: Seems this is only used for a no-op logger (unless I missed something). Not sure it's worth the dependency, were you planning on keeping this or is it just temporary?

websocket/config.go
```diff
@@ -0,0 +1,42 @@
+package polygonws
+
+import "time"
+
+// todo: add more config options with validation (waits, feed/market, etc)
```
mmoghaddam385, line 5: I forgot to mention this the other day, but I wanna suggest we support a config option that doesn't parse messages (except for maybe status messages). Something like this in the jvm client: https://github.com/polygon-io/client-jvm/blob/master/src/main/kotlin/io/polygon/kotlin/sdk/websocket/PolygonWebSocketClient.kt#L61-L64

My main motivation for including that in the jvm client was so if for whatever reason the models in this client get out of sync with what the websocket actually sends, users will have the ability to parse messages using their own models.

Not saying add that to this PR now or anything, just wanted to share a thought

mmoghaddam385 (commented):
I really like the direction here! The structure makes a lot of sense so far, hopefully once we add more functionality it stays this clear 🤞

websocket/polygon.go
morningvera: makes sense! i'll run through the existing code and clean things up a bit

websocket/polygon.go
morningvera: it's a bit unclear to me but i _think_ calling WriteControl concurrently with WriteMessage is allowed (see this [issue](https://github.com/gorilla/websocket/issues/83#issuecomment-147906076)). old issue but i've seen this kind of pattern elsewhere too. going to dig some more and regardless, might refactor this to use the wQueue instead

go.mod
morningvera: good point! going to implement a no-op logger instead of using this

websocket/polygon.go
morningvera: leaving a todo comment here so i'll remember to verify later, going to merge as is for now

websocket/config.go
morningvera: i like it! left a todo comment about it. i'll figure out a good way to handle it as we get subscriptions and data handling in place

websocket/polygon.go
morningvera: leaving things as is for now (it's helpful for developing) but i left a comment to clean things up as we get more stable

mcdayoub (approved):

websocket/polygon.go
```diff
@@ -158,46 +157,41 @@ func (c *Client) process() {
...
-
-	return nil
 }
 
-func (c *Client) handleStatus(msg json.RawMessage) error {
+func (c *Client) handleStatus(msg json.RawMessage) {
 	var cm models.ControlMessage
 	if err := json.Unmarshal(msg, &cm); err != nil {
-		return err
+		c.log.Errorf("failed to unmarshal message")
```
mmoghaddam385, line 193: This should log the error message since it's no longer being returned

mmoghaddam385 (approved):

PR #72, opened 2022-04-18
Added subscriptions set to client struct 

Darcy-Linde:


websocket/polygon.go
```diff
@@ -125,12 +129,35 @@ func getParams(market Market, topic Topic, tickers ...string) (string, error) {
...
+	}
+	c.subscriptions[prefix][ticker] = struct{}{}
+}
+
+func (c *Client) deleteSubscription(prefix string, ticker string) {
+	_, exists := c.subscriptions[prefix]
+	if !exists {
+		c.subscriptions[prefix] = make(set)
+	}
+	if _, includes := c.subscriptions[prefix][ticker]; includes {
```
mcdayoub, line 145: bang

mcdayoub (approved):

mcdayoub (approved):

websocket/polygon.go
```diff
@@ -125,12 +129,34 @@ func getParams(market Market, topic Topic, tickers ...string) (string, error) {
 	return strings.Join(params, ","), nil
 }
 
+func (c *Client) setSubscription(prefix string, ticker string) {
```
morningvera, line 132: i think these should've used variadic args: `setSubscription(topic Topic, tickers ...string)`

Darcy-Linde: I could change the implementation, but as it stands setSubscription only takes in one ticker at a time so we wouldn't need tickers ...string. As for my use of prefix over topic, I did that because the parameter is taking in topic.prefix(). I thought it would be better to be more specific.

morningvera: i'm suggesting that you should change the implementation. topic is more descriptive than prefix imo (you subscribe to a topic, not a prefix).

morningvera: no, the helper is fine but i don't think you should force users of this func to loop over variadic args when they could pass them to the helper instead (similar to `getParams(...)`)

PR #73, opened 2022-04-18
clean up close logic

morningvera:
moving the close logic into the write thread. it's more thread-safe and prevents us from writing to the conn after close.

### DIFF
--- a/websocket/polygon.go
+++ b/websocket/polygon.go
@@ -14,6 +14,7 @@ import (
 
 // todo: add reconnect logic
 // todo: in general, successful calls should be debug and unknown messages should be info
+// todo: probably remove some junk logging before release too
 
 type Client struct {
 	apiKey string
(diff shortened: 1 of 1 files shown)


websocket/polygon.go
```diff
@@ -195,7 +187,6 @@ func (c *Client) authenticate() error {
 func (c *Client) read() {
 	defer func() {
 		c.log.Debugf("closing read thread")
-		c.conn.Close() // todo: should this force close?
```
morningvera, line 198: only the write thread should force close

websocket/polygon.go
```diff
@@ -224,12 +215,18 @@ func (c *Client) write() {
 	defer func() {
 		c.log.Debugf("closing write thread")
 		ticker.Stop()
-		c.conn.Close() // todo: should this force close?
+		c.conn.Close()
```
morningvera, line 218: this is safe now that this is the only thread that closes the connection

websocket/polygon.go
```diff
@@ -240,7 +237,7 @@ func (c *Client) write() {
 		case msg := <-c.wQueue:
 			if err := c.conn.SetWriteDeadline(time.Now().Add(writeWait)); err != nil {
 				c.log.Errorf("failed to set write deadline: %v", err)
-				return // todo: should this return?
+				return
```
morningvera, line 240: yes, assuming we'll just reconnect

mcdayoub (approved):

PR #76, opened 2022-04-19
refactored sub/unsub helpers

Darcy-Linde:


websocket/polygon.go
```diff
@@ -194,11 +193,20 @@ func (c *Client) Unsubscribe(topic Topic, tickers ...string) error {
 	return nil
 }
 
-func (c *Client) Close() {
+func (c *Client) Close() error {
```
morningvera, line 196: this all seems like a bad rebase. it's reverting some changes i made yesterday.

Darcy-Linde: When I made my branch I thought I had branched from master but I think I had previously set my origin to point to a different branch. I just quickly fixed it in a new branch and I'll push that instead of rebasing this.

PR #80, opened 2022-04-20
added unit tests and moved live tests into separate file.

Darcy-Linde:


websocket/polygon_test.go
```diff
@@ -1,99 +1,57 @@
...
+	_, aapl = c.subscriptions["AM"]["AAPL"]
+	assert.Equal(t, false, aapl)
+	c.setSubscriptions(Topic(12), "*")
+	_, all := c.subscriptions["AM"]["*"]
+	_, tsla = c.subscriptions["AM"]["TSLA"]
+	assert.Equal(t, false, tsla)
+	assert.Equal(t, true, all)
+	_, trade := c.subscriptions["T"]
+	assert.Equal(t, false, trade)
+	c.deleteSubscriptions(Topic(13), "RDFN")
```
mcdayoub, line 54: Is there anything we can assert if there is a delete here? I get that `RDFN` did not exist. 

Maybe it means we need to log that it wasn't in the subscriptions set, and for the test we could possibly get the length of the set before the delete and then assert that the length does not change after the delete

Darcy-Linde: Good point. I'll add a test for that.

Darcy-Linde: Actually in this case, the set does not exist before line 54. This is testing the edge case where a user tries to delete from a topic that hasn't been instantiated yet. This is why I run assert.Equal(t, true, trade) afterwards to make sure that now c.subscriptions["T"] has been created.

mcdayoub: oh that makes sense, nice

mcdayoub (approved):

websocket/polygon_test.go
```diff
@@ -1,99 +1,57 @@
...
-	}
-
-	for {
-		select {
-		case <-ctx.Done():
-			return
-		}
-	}
+func TestSupportsTopic(t *testing.T) {
+	assert.Equal(t, true, supportsTopic("stocks", Topic(12)))
```
morningvera, line 11: can we use the actual topic var instead of casting an int to Topic? this isn't very readable as is

websocket/polygon_test.go
```diff
@@ -1,99 +1,57 @@
...
-func printOutput(c *polygonws.Client) {
-	for {
-		out := c.Output()
-		if out == nil {
-			break
-		}
-		fmt.Println(out)
-	}
+func TestGetParams(t *testing.T) {
+	p1, _ := getParams("stocks", Topic(12), "AAPL", "GME", "HOOD")
```
morningvera, line 23: same here (and anywhere else)

websocket/polygon_test.go
```diff
@@ -1,99 +1,57 @@
...
+	p2, _ := getParams("stocks", Topic(12))
+	s1 := "AM.AAPL,AM.GME,AM.HOOD"
+	s2 := "AM.*"
+	assert.Equal(t, p1, s1)
+	assert.Equal(t, p2, s2)
 }
+func TestSubscriptions(t *testing.T) {
```
morningvera, line 30: can we add a test case for deleting `*`? are we handling that properly?

morningvera (approved):

PR #84, opened 2022-04-22
Darcy.refactor sub unsub

Darcy-Linde:


websocket/polygon.go
```diff
@@ -128,7 +128,18 @@ func (c *Client) connect() error {
 	return nil
 }
 
+func (c *Client) checkPrefix(topic Topic) {
```
morningvera, line 131: given the name, i'd expect this interface to look like `checkPrefix(topic Topic) bool` but i get that this is actually supposed to set missing topic prefixes. maybe `setTopic` or `setPrefix` would be more clear?

websocket/polygon.go
```diff
@@ -128,7 +128,18 @@ func (c *Client) connect() error {
...
+	if _, prefixExists := c.subscriptions[topic.prefix()]; !prefixExists {
+		c.subscriptions[topic.prefix()] = make(set)
+	}
+}
+
 func (c *Client) Subscribe(topic Topic, tickers ...string) error {
```
morningvera, line 137: i don't think this supports the `Subscribe(StocksSecAggs)` case anymore. if no tickers are passed, we should assume the user means `Subscribe(StocksSecAggs, "*")`. probably worth adding this to the live test as well (or modifying the part that you added). Unsubscribe should do the same thing

websocket/polygon.go
```diff
@@ -448,11 +464,10 @@ func getParams(market Market, topic Topic, tickers ...string) (string, error) {
 }
 
 func (c *Client) setSubscriptions(topic Topic, tickers ...string) {
+	if tickers[0] == "*" {
```
morningvera, line 467: this could panic if `len(tickers) == 0` btw. not that much of an issue since it's internal logic but might want to guard against it anyway (i.e. `if len(tickers) > 0 && tickers[0] == "*"`)

morningvera (changes requested):

morningvera (approved):

PR #89, opened 2022-04-25
add automatic reconnects

morningvera:
steps to test the reconnect logic:

1. turn off wifi
2. run the live test (it should retry until wi-fi is turned back on)
3. turn on wifi
4. watch some messages come in
5. turn off wifi again (it should go back to the retry loop)
6. turn on wifi
7. watch more messages come in 🙂

pls try this and attempt to break it before approving

### DIFF
--- a/websocket/polygon.go
+++ b/websocket/polygon.go
@@ -21,6 +21,7 @@ import (
 
 type set map[string]struct{}
 
+// Client defines a client to the Polygon WebSocket API.
 type Client struct {
 	apiKey        string
 	feed          Feed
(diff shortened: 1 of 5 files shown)


websocket/polygon.go
```diff
@@ -45,11 +45,7 @@ type Client struct {
 	log Logger
 }
 
-func New(ctx context.Context, config Config) (*Client, error) {
-	if ctx == nil {
-		ctx = context.Background()
-	}
-
+func New(config Config) (*Client, error) {
```
morningvera, line 48: i don't think we actually need this context here. users should call Close instead of cancelling via context.

websocket/polygon.go
```diff
@@ -136,6 +165,9 @@ func (c *Client) setTopic(topic Topic) {
 }
 
 func (c *Client) Subscribe(topic Topic, tickers ...string) error {
+	c.mtx.Lock()
```
morningvera, line 168: need to guard these methods since we push to the write queue (which could be cleared during reconnects)

mcdayoub (approved):

PR #92, opened 2022-04-27
define a type for subscriptions

morningvera:
attempting to move some logic out of the main client so we can unit test it more easily. also added some more test coverage. i thought about making this its own package but i don't want to export it.


Darcy-Linde (approved):
I like this abstraction of subscriptions. Looks a lot cleaner now

PR #97, opened 2022-04-28
clean up client creation and startup

morningvera:
Added some test coverage around connect/auth but it's difficult to actually verify it within the test. Three of the test cases output logs though so we can at least visually verify them. Thinking we'll improve these over time.


mcdayoub (approved):
great job on this

PR #99, opened 2022-04-29
update readme and godoc

morningvera:


Darcy-Linde (approved):

websocket/polygon.go
```diff
@@ -154,41 +154,23 @@ func (c *Client) Unsubscribe(topic Topic, tickers ...string) error {
 	return nil
 }
 
+// Output returns the next message in the output queue. If no messages are available, it returns nil.
 func (c *Client) Output() any {
-	return <-c.output
```
morningvera, line 158: this was a blocking call which could cause users to deadlock if the client isn't receiving data

websocket/polygon_test.go
```diff
@@ -90,6 +90,10 @@ func TestConnectAuthSuccess(t *testing.T) {
 	// closing before connecting shouldn't do anthing
 	c.Close()
 
+	// accessing output early shouldn't do anything
```
calebcase, line 93: I wish these were full sentences :)

calebcase (approved):

README.md
```diff
@@ -91,6 +91,86 @@ if err != nil {
...
+
+Next, create a new client with your API key and a few other config options.
+
+```golang
+// create a new client
+c, err := polygonws.New(polygonws.Config{
+    APIKey:    "YOUR_API_KEY",
+    Feed:      polygonws.RealTime,
+    Market:    polygonws.Stocks,
+    ParseData: true, // this tells the client to unmarshal data into models
```
qrpike, line 113: I think `ParseData` is going to be 99% of clients. Not sure the use case for this, really. But, It would be nice to not require fields ( requiring users time, brainpower ) for majority of use cases.

morningvera: that's a fair point! i could flip it to something like `RawData` so users have to intentionally disable parsing data

websocket/config.go
```diff
@@ -4,13 +4,29 @@ import (
 	"errors"
 )
 
+// Config is a set of WebSocket client options.
 type Config struct {
-	APIKey     string
-	Feed       Feed
-	Market     Market
+	// APIKey is the API key used to authenticate against the server.
+	APIKey string
```
qrpike, line 10: Should this be private for security purposes? Is there any need for external use of it?

morningvera: it needs to be public so users can pass it to us. i don't think there are any security implications to setting a field here

README.md
```diff
@@ -91,6 +91,86 @@ if err != nil {
...
+}
+defer c.Close() // the user of this client must close it
+
+// connect to the server
+if err := c.Connect(); err != nil {
+    log.Fatal(err)
+}
+```
+
+The client automatically reconnects to the server when the connection is dropped. By default, it will attempt to reconnect indefinitely but the number of retries is configurable. When the client successfully reconnects, it automatically resubscribes to any topics that were set before the disconnect.
```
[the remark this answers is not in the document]
qrpike: 🙌

websocket/config.go
qrpike: Never-mind, this is the config struct. Disregard.

websocket/polygon_test.go
morningvera: what specifically are you looking for? seems clear enough to me

websocket/polygon.go
```diff
@@ -154,41 +154,23 @@ func (c *Client) Unsubscribe(topic Topic, tickers ...string) error {
...
+	select {
+	case out := <-c.output:
```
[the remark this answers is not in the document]
morningvera: good catch! updating it

websocket/polygon_test.go
morningvera: oh, that's not how i generally do inline comments. happy to change it if this is a polygon specific style but probably not in this PR.

calebcase: It's a style I promote. When I see comments like this I think things like "these comments don't matter because the writer didn't care enough to make them full sentences" or "these comments will be low quality because..." or "the writer didn't care about the reader needing to read it so I probably don't need to read it"

morningvera: hmm, i'm not sure i agree but i'll consider changing it in a future PR. for now, i don't want to update it in this one specific place when the rest of my inline comments match my style

PR #103, opened 2022-05-02
expose output channel and guard against user closing it

morningvera:
The original design was causing this client to be CPU hungry. Attempting to expose the channel in a relatively safe way. There's no way to stop the user from closing it but the client should still recover the panic and close gracefully.

### DIFF
--- a/websocket/polygon.go
+++ b/websocket/polygon.go
@@ -154,17 +154,10 @@ func (c *Client) Unsubscribe(topic Topic, tickers ...string) error {
 	return nil
 }
 
-// Output returns the next message in the output queue. If no messages are available, it returns nil.
-func (c *Client) Output() any {
-	select {
-	case out, ok := <-c.output:
-		if ok {
-			return out
-		}
-		return nil
-	default:
-		return nil
-	}
+// Output returns the output queue. If the channel is closed by the user (not
+// recommended), the client connection will close as well.
+func (c *Client) Output() chan any {
+	return c.output
 }
 
 // Close attempt to gracefully close the connection to the server.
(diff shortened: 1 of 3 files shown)


websocket/polygon.go
```diff
@@ -329,16 +329,15 @@ func (c *Client) write() error {
 			if err := c.conn.WriteMessage(websocket.TextMessage, msg); err != nil {
 				return fmt.Errorf("failed to send message: %w", err)
 			}
-		default:
```
morningvera, line 332: i think these defaults could cause the write and process threads to be cpu hungry. deleting so they can block until data is available.

websocket/polygon_live_test.go
```diff
@@ -91,35 +93,17 @@ func TestMain(t *testing.T) {
...
-			return
-		default:
-			out := client.Output()
-			if out == nil {
-				continue
-			}
-			fmt.Println(out)
-		}
+func printOutput(client *polygonws.Client) {
+	for out := range client.Output() {
```
morningvera, line 97: this looks way cleaner. much happier with it.

websocket/polygon.go
```diff
@@ -269,8 +262,15 @@ func (c *Client) close(reconnect bool) {
...
+
 		c.shouldClose = true
-		close(c.output)
+
+		select {
+		case <-c.output:
+			c.log.Debugf("output channel was closed by user")
+		default:
+			close(c.output)
+		}
```
mmoghaddam385, line 273: If a user hasn't read a message off of the output channel would this end up not closing the output channel?

morningvera: i was actually worried about that as well. tried testing that scenario and the channel seemed to close but now i'm skeptical. looking into it a bit more

morningvera: just confirmed that you're right. i added a helper that closes and recovers from the panic 🙂

mmoghaddam385 (approved):

PR #104, opened 2022-05-03
Stock Financials Endpoint

Darcy-Linde:


rest/reference.go
```diff
@@ -128,6 +130,14 @@ func (c *ReferenceClient) ListDividends(ctx context.Context, params *models.List
 	})
 }
 
+func (c *ReferenceClient) ListFinancials(ctx context.Context, params *models.ListFinancialsParams, options ...models.RequestOption) *iter.Iter[models.Financial] {
```
morningvera, line 133: we should separate this into its own `VXClient`. i want it to be super clear that users are using something that could have breaking changes. this method also needs to be documented and we should mention that breaking changes (or deprecation) could happen on it

morningvera (changes requested):

rest/reference.go
morningvera: i also don't want the top-level client to inherit the `VXClient` so instead of users calling `c.ListFinancials`, i want them to call `c.VX.ListFinancials`

Darcy-Linde: Would you like a separate financials.go file, or should I just make a second client in the reference.go file?

morningvera: it should probably be a vx.go file or something along those lines

rest/polygon.go
```diff
@@ -13,6 +13,7 @@ type Client struct {
 	ReferenceClient
 	TradesClient
 	SnapshotClient
+	VXClient
```
morningvera, line 16: this needs to be `VX VXClient`. what you're doing here is go's weird form of inheritance which i don't want in this case

rest/models/financials.go
```diff
@@ -0,0 +1,132 @@
...
+	Results []Financial `json:"results,omitempty"`
+}
+
+// Financial contains detailed information on a specified stock financial.
+type Financial struct {
+	CIK                 string                 `json:"cik,omitempty"`
+	CompanyName         string                 `json:"company_name,omitempty"`
+	EndDate             string                 `json:"end_date,omitempty"`
+	FilingDate          string                 `json:"filing_date,omitempty"`
+	Financials          map[string]interface{} `json:"financials,omitempty"`
```
morningvera, line 126: this type is likely going to be hard to use. we should define a struct instead of using `interface{}`

rest/models/types.go
```diff
@@ -136,6 +136,22 @@ const (
...
+const (
+	FULL   NameComparator = "full"
+	SEARCH NameComparator = "search"
+)
+
+// TimeFrame is the type of time frame query parameter for stock financials.
+type Timeframe string
+
+const (
+	ANNUAL    Timeframe = "annual"
```
morningvera, line 151: these should be camelcase like the rest of our enums. only acronyms should be all caps (like EQ, LTE, etc)

morningvera (changes requested):

rest/models/types.go
Darcy-Linde: We have two const declarations of Quarterly (one in Frequency, and one here in TimeFrame). Is there a convention for dealing with this?

morningvera: yeah, my convention is to just use the existing Frequency enum instead of adding a new Timeframe enum (it seems like they're functionally the same in this case)

rest/models/financials.go
```diff
@@ -0,0 +1,141 @@
...
+	FilingDate          string               `json:"filing_date,omitempty"`
+	Financials          map[string]Financial `json:"financials,omitempty"`
+	FiscalPeriod        string               `json:"fiscal_period,omitempty"`
+	FiscalYear          string               `json:"fiscal_year,omitempty"`
+	SourceFilingFileUrl string               `json:"source_filing_file_url,omitempty"`
+	SourceFilingUrl     string               `json:"source_filing_url,omitempty"`
+	StartDate           string               `json:"start_date,omitempty"`
+}
+
+type Financial map[string]struct {
```
morningvera, line 134: can you add a godoc comment for this?

morningvera (approved):

PR #110, opened 2022-05-11
add version number to user agent header

morningvera:
Couldn't think of an ideal way to automate this so I ended up writing a script. Currently, only admins can run it.

### DIFF
--- a/rest/client/client.go
+++ b/rest/client/client.go
@@ -31,7 +31,7 @@ func New(apiKey string) Client {
 	c.SetAuthToken(apiKey)
 	c.SetRetryCount(DefaultRetryCount)
 	c.SetTimeout(10 * time.Second)
-	c.SetHeader("User-Agent", fmt.Sprintf("Go client %v", clientVersion))
+	c.SetHeader("User-Agent", fmt.Sprintf("Polygon.io GoClient/%v", clientVersion))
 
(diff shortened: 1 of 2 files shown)


jbonzo (approved):

rest/client/client.go
```diff
@@ -29,7 +31,7 @@ func New(apiKey string) Client {
 	c.SetAuthToken(apiKey)
 	c.SetRetryCount(DefaultRetryCount)
 	c.SetTimeout(10 * time.Second)
-	c.SetHeader("User-Agent", "Go client")
+	c.SetHeader("User-Agent", fmt.Sprintf("Go client %v", clientVersion))
```
jbonzo, line 34: before major release we will want a better user agent that includes `Polygon` or something. And have it be standard across the other client libs

morningvera: agreed, maybe we should change that now? `Polygon Go client v0.7.1`?

jbonzo: `Polygon.io GoClient/v0.7.1`
or 
`Polygon.io GoClient/0.7.1`

morningvera: i think i'll go with the first one so it matches the repo tags

jbonzo (approved):

PR #113, opened 2022-05-13
Rest Example based on README

Darcy-Linde:


rest/example/main.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"context"
+	"log"
+	"os"
+
+	polygon "github.com/polygon-io/client-go/rest"
+	"github.com/polygon-io/client-go/rest/models"
+)
+
+// See README for more examples.
```
morningvera, line 12: probably don't need this comment

morningvera (approved):
could you link to this example in the readme? similar to [this](https://github.com/polygon-io/client-go/blob/master/README.md?plain=1#L164)

rest/example/main.go
Darcy-Linde: Ok I'll take it out

PR #115, opened 2022-05-13
document the rest of the reference api

morningvera:


rest/models/conditions.go
```diff
@@ -55,15 +67,35 @@ type ListConditionsResponse struct {
...
-	AssetClass   string                     `json:"asset_class,omitempty"`
-	DataTypes    []string                   `json:"data_types,omitempty"`
-	Description  string                     `json:"description,omitempty"`
-	Exchange     int64                      `json:"exchange,omitempty"`
-	ID           int64                      `json:"id,omitempty"`
-	Legacy       bool                       `json:"legacy"`
-	Name         string                     `json:"name,omitempty"`
-	SIPMapping   map[string]string          `json:"sip_mapping,omitempty"`
-	Type         string                     `json:"type,omitempty"`
-	UpdateRules  map[string]map[string]bool `json:"update_rules,omitempty"`
```
morningvera, line 68: not sure why we didn't define a struct here originally

rest/models/tickers.go
```diff
@@ -267,71 +267,62 @@ type GetTickerTypesResponse struct {
...
-	PostalCode string `json:"postal_code,omitempty"`
-	State      string `json:"state,omitempty"`
-}
-
-// Branding contains information related to a company's brand.
-type Branding struct {
-	LogoURL string `json:"logo_url,omitempty"`
-	IconURL string `json:"icon_url,omitempty"`
+	Active  bool `json:"active"`
+	Address struct {
```
morningvera, line 271: just seeing how people feel about this. i think i like it but i'm going to think about it over the weekend before merging.

Darcy-Linde: Were we defining the children structs outside of the parent struct before and now they've been moved inside? (Or were we not defining them at all before?)

PR #120, opened 2022-05-16
Option Contracts Endpoints

Darcy-Linde:


rest/models/contracts.go
```diff
@@ -0,0 +1,161 @@
+package models
+
+// GetOptionsContract is the set of parameters for the GetOptionsContract method.
+type GetOptionsContractParams struct {
+	// Return the contract that contains this options ticker.
+	OptionsTicker string `validate:"required" path:"options_ticker"`
```
morningvera, line 6: I think we could change this to follow our typical pattern without losing clarity:
```
Ticker string `validate:"required" path:"ticker"`
```

rest/models/contracts.go
```diff
@@ -0,0 +1,161 @@
...
+
+	// Specify a point in time for the contract as of this date.
+	AsOf *Date `query:"as_of"`
+}
+
+func (p GetOptionsContractParams) WithDate(q Date) *GetOptionsContractParams {
```
morningvera, line 12: `WithDate` -> `WithAsOf`

rest/reference.go
```diff
@@ -25,6 +25,10 @@ const (
 	ListConditionsPath = "/v3/reference/conditions"
 
 	GetExchangesPath = "/v3/reference/exchanges"
+
+	GetOptionsContractPath = "/v3/reference/options/contracts/{options_ticker}"
+
```
morningvera, line 30: I've been grouping these paths intentionally so we should get rid of this extra space here

rest/reference.go
```diff
@@ -155,3 +159,19 @@ func (c *ReferenceClient) GetExchanges(ctx context.Context, params *models.GetEx
 	err := c.Call(ctx, http.MethodGet, GetExchangesPath, params, res, options...)
 	return res, err
 }
+
+// GetOptionsContract retrieves a historical options contract.
```
morningvera, line 163: Add links to the docs pls (see any other method for examples)

rest/reference.go
```diff
@@ -155,3 +159,19 @@ func (c *ReferenceClient) GetExchanges(ctx context.Context, params *models.GetEx
...
+func (c *ReferenceClient) GetOptionsContract(ctx context.Context, params *models.GetOptionsContractParams, options ...models.RequestOption) (*models.GetOptionsContractResponse, error) {
+	res := &models.GetOptionsContractResponse{}
+	err := c.Call(ctx, http.MethodGet, GetOptionsContractPath, params, res, options...)
+	return res, err
+}
+
+// ListOptionsContracts lists historical options contracts.
```
morningvera, line 170: Also add the iter example to this comment (see any other list method)

rest/models/contracts.go
```diff
@@ -0,0 +1,161 @@
...
+	p.AsOf = &q
+	return &p
+}
+
+// GetOptionsContractResponse is the response returned by the GetOptionsContract method.
+type GetOptionsContractResponse struct {
+	BaseResponse
+
+	// Options Contract results.
+	Results OptionsContract `json:"results,omitempty"`
```
morningvera, line 22: i decided not to document the response fields right now so we should remove this (and the line above) to be consistent

rest/models/contracts.go
```diff
@@ -0,0 +1,161 @@
...
+func (p ListOptionsContractsParams) WithLimit(q int) *ListOptionsContractsParams {
+	p.Limit = &q
+	return &p
+}
+
+type ListOptionsContractsResponse struct {
+	BaseResponse
+
+	// Options Contracts results.
+	Results []OptionsContract `json:"results,omitempty"`
```
morningvera, line 138: same comment here

morningvera (changes requested):
looks good! just some minor things

rest/models/contracts.go
Darcy-Linde: Changed OptionsTicker to Ticker, but the path is options_ticker so I left that as is. 
/v3/reference/options/contracts/{options_ticker}

rest/reference.go
```diff
@@ -149,9 +152,37 @@ func (c *ReferenceClient) ListConditions(ctx context.Context, params *models.Lis
 	})
 }
 
-// GetExchanges lists all exchanges that Polygon knows about.
+// GetExchanges lists all exchanges that Polygon knows about. For more details see
+// https://polygon.io/docs/stocks/get_v3_reference_exchanges
```
morningvera, line 156: thanks for catching this one! still need to add a `.` after all of these links

PR #150, opened 2022-06-17
make querying a whole day easier for list trades/quotes

morningvera:
I think this solution makes sense and shouldn't really break anyone even though it changes internal behavior slightly. Try `POLYGON_API_KEY=<your-api-key> make rest-example` if you want to see it working


jbonzo (commented):
Ugh... I hate to nitpick but what if I wanted trades that only happened on the zeroed out values.

morningvera:
@jbonzo ahh, i totally meant to use `&&`. that's not a nitpick, good catch!

jbonzo:
> @jbonzo ahh, i totally meant to use `&&`. that's not a nitpick, good catch!

mmm no my question still stands (i didnt even notice that bug)

If i wanted trades that happened exactly on `2006-01-02T15:00:00.000000001Z07:00` then I would get only the trades that happened on that nanosecond, but if I asked for `2006-01-02T15:00:00.000000000Z07:00` then I would get the entire day regardless of my intention.

morningvera:
> @jbonzo
> If i wanted trades that happened exactly on `2006-01-02T15:00:00.000000001Z07:00` then I would get only the trades that happened on that nanosecond, but if I asked for `2006-01-02T15:00:00.000000000Z07:00` then I would get the entire day regardless of my intention.

The `isDay` helper checks for the hour as well so I don't think that's an issue?

morningvera:
refactored this and added a `WithDay(...)` method to the param builders

morningvera:
alright, i think the encoding solution is what we're going with. i also included a `WithDay` helper to make it easier to use.

rest/encoder/encoder.go
```diff
@@ -96,8 +96,20 @@ func newEncoder(tag string) *form.Encoder {
...
 		return []string{fmt.Sprint(time.Time(x.(models.Nanos)).UnixNano())}, nil
 	}, models.Nanos{})
 
 	return e
 }
+
+func isDay(t time.Time) bool {
+	if t.Hour() != 0 || t.Minute() != 0 || t.Second() != 0 || t.Nanosecond() != 0 {
+		return false
+	}
```
joedursun, line 113: Would it be more robust if we truncated to the day and compared like so?

```golang
return t.Truncate(24*time.Hour) == t
```

Here's a [playground](https://play.golang.com/p/PAZRKkZHk8y) demonstrating it.

morningvera: Thanks for the example! Currently, this PR is assuming users want a day if they give us midnight in any timezone (which might be an incorrect assumption). Like for example, does a user want a whole day of crypto trades if they're asking for `2021-07-22 00:00:00 UTC+3`? Or do we only assume this if they use UTC?

Also, this comment just made me realize that we probably can't go with this solution at all. I'm pretty sure if a user wants to do something like `params.TimestampGTE = 1626946200000000000` (2021-07-22 09:30:00) and `params.TimestampLTE = 1626969600000000000` (2021-07-23 16:00:00), it'll be marshaled to `timestamp.gte=2021-7-22&timestamp.lte=2021-7-22` 😭 I can't think of a way to only encode this way for `timestamp.eq`.

joedursun (approved):

PR #194, opened 2022-10-03
update interator pattern fix

### LINKED ISSUE #191: Examples code seems incorrect
**Describe the bug**
The example code for some endpoints have invalid code.

**To Reproduce**
https://github.com/polygon-io/client-go/blob/master/rest/reference.go#L104-L110

**Expected behavior**
```
//	iter := c.ListSplits(context.TODO(), params, opts...)
//	for iter.Next() {
//	    log.Print(iter.Item()) // do something with the current value
//	}
//	if iter.Err() != nil {
//	    return iter.Err()
//	}
```

**Current Behavior**
```
//	iter, err := c.ListSplits(context.TODO(), params, opts...)
//	for iter.Next() {
//	    log.Print(iter.Item()) // do something with the current value
//	}
//	if iter.Err() != nil {
//	    return err
//	}
```

Darcy-Linde:
Closes #191

### DIFF
--- a/rest/reference.go
+++ b/rest/reference.go
@@ -42,7 +42,7 @@ type ReferenceClient struct {
 //
 //	iter := c.ListTickers(context.TODO(), params, opts...)
 //	for iter.Next() {
-//	    log.Print(iter.Item()) // do something with the current value
+//		log.Print(iter.Item()) // do something with the current value
 //	}
 //	if iter.Err() != nil {
 //		return iter.Err()
@@ -70,7 +70,7 @@ func (c *ReferenceClient) GetTickerDetails(ctx context.Context, params *models.G
 //
 //	iter := c.ListTickerNews(context.TODO(), params, opts...)
 //	for iter.Next() {
-//	    log.Print(iter.Item()) // do something with the current value
+//		log.Print(iter.Item()) // do something with the current value
 //	}
 //	if iter.Err() != nil {
 //		return iter.Err()
--- a/rest/reference.go
+++ b/rest/reference.go
@@ -187,7 +187,7 @@ func (c *ReferenceClient) GetOptionsContract(ctx context.Context, params *models
 //
 // This method returns an iterator that should be used to access the results via this pattern:
 //
-//	iter := c.ListConditions(context.TODO(), params, opts...)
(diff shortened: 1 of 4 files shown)


rest/reference.go
```diff
@@ -189,10 +180,7 @@ func (c *ReferenceClient) GetOptionsContract(ctx context.Context, params *models
 //
 // This method returns an iterator that should be used to access the results via this pattern:
 //
-//	iter, err := c.ListConditions(context.TODO(), params, opts...)
-//	if err != nil {
-//		return err
-//	}
+//	iter := c.ListConditions(context.TODO(), params, opts...)
```
morningvera, line 183: looks like this is a separate issue but can you fix it? `c.ListConditions` -> `c.ListOptionsContract`

rest/vx.go
```diff
@@ -25,10 +25,7 @@ type VXClient struct {
 //
 // This method returns an iterator that should be used to access the results via this pattern:
 //
-//	iter, err := c.ListStockFinancials(context.TODO(), params, opts...)
-//	if err != nil {
-//		return err
-//	}
+//	iter := c.ListStockFinancials(context.TODO(), params, opts...)
 //	for iter.Next() {
 //	    log.Print(iter.Item()) // do something with the current value
```
morningvera, line 30: why is the spacing changed on some of them but not this one?

Darcy-Linde: They look the same in my editor, but some comments were done with tabs and some were done with spaces it seems. Running `go fmt` didn't fix it so I just made sure that they were all indented with a tab. Let me know if they still don't look right after my last change.

morningvera: looks good now, thanks!

morningvera (approved):
lgtm 🎉

rest/reference.go
```diff
@@ -104,10 +111,7 @@ func (c *ReferenceClient) GetMarketStatus(ctx context.Context, options ...models
 //
 // This method returns an iterator that should be used to access the results via this pattern:
 //
-//	iter, err := c.ListSplits(context.TODO(), params, opts...)
-//	if err != nil {
-//		return err
-//	 }
```
qrpike, line 110: If we are removing the error handling, we should add back the correct way of checking for errors.

PR #196, opened 2022-10-09
fix: use correct wording

nolotz:
Hi there,

this fixes a wording issue in CryptoTrade struct.

https://polygon.io/docs/crypto/ws_crypto_xt

### DIFF
--- a/websocket/models/models.go
+++ b/websocket/models/models.go
@@ -147,7 +147,7 @@ type CryptoTrade struct {
 	EventType
 
 	// The crypto pair.
-	Symbol string `json:"sym,omitempty"`
+	Pair string `json:"pair,omitempty"`
 
 	// The crypto exchange ID.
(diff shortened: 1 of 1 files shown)


morningvera:
@nolotz thanks for the PR! We generally expect outside contributors to open issues before PRs (see [contributing](https://github.com/polygon-io/client-go#contributing)). I went ahead and wrote an [issue](https://github.com/polygon-io/client-go#contributing) for this. In this case, what you're proposing is a breaking change so we can't merge it. I'm going to decline this PR and work on a backwards compatible fix but thanks again for contributing!

nolotz:
@morningvera Good Idea! Thank you for owning this.

nolotz:
For those who have the same issue distinguishing between individual Crypto Trades: 
You can still set RawData in your polygonws.Config to true and parse it with a custom struct & json implementation.

morningvera (approved):

PR #198, opened 2022-10-10
Change split_from and split_to from int64 to float64

clickingbuttons:
Bugfix for this split:
```
2022/10/10 13:26:39.854395 ERROR RESTY json: cannot unmarshal number 3.83227428463e+11 into Go struct field Split.results.split_from of type int64, Attempt 1
Error: failed to execute request: json: cannot unmarshal number 3.83227428463e+11 into Go struct field Split.results.split_from of type int64
```

To match the models in [go-apps-reference-splits](https://github.com/polygon-io/go-apps-reference-splits/blob/master/models/models.go#L6).

### DIFF
--- a/rest/models/splits.go
+++ b/rest/models/splits.go
@@ -91,7 +91,7 @@ type ListSplitsResponse struct {
 // Split contains detailed information on a specified stock split.
 type Split struct {
 	ExecutionDate string `json:"execution_date,omitempty"`
-	SplitFrom     int64  `json:"split_from,omitempty"`
-	SplitTo       int64  `json:"split_to,omitempty"`
+	SplitFrom     float64  `json:"split_from,omitempty"`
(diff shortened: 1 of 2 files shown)


qrpike:
@Darcy-Linde I believe you did these models. Can you please check the other models closely to make sure they match the docs.

qrpike (approved):
@morningvera your opinions around versioning? Patch vs Minor?

morningvera:
@qrpike @clickingbuttons isn't this technically of a breaking change?

qrpike:
> @qrpike @clickingbuttons isn't this technically a breaking change?

Yes, but it's also a critical bug. That's why I'm curious what you think it should be released as.

clickingbuttons:
I need to run some Docker things that have this change so I'm merging but not releasing.

PR #202, opened 2022-10-18
Set issue deadline of 3w

clickingbuttons:
Same as https://github.com/polygon-io/client-python/pull/318


morningvera (approved):
thanks for doing this one too!

PR #216, opened 2022-11-29
Launchpad required headers

jbonzo:
Addresses https://app.zenhub.com/workspaces/backend-6042935fc77665000f50780e/issues/polygon-io/backend/749

Work in progress.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,41 +118,7 @@ log.Print(res) // do something with the result
 #### Launchpad usage
 
 Users of the Launchpad product will need to pass in certain headers in order to make API requests.
-
-```golang
-params := &models.GetGroupedDailyAggsParams{
-    Locale:     models.US,
-    MarketType: models.Stocks,
-    Date:       models.Date(time.Date(2021, 7, 22, 0, 0, 0, 0, time.Local)),
-}
-
-res, err := c.GetGroupedDailyAggs(context.Background(), params,
-    models.RequiredEdgeHeaders("EDGE_USER_ID", "EDGE_USER_IP_ADDRESS"),
-)
-if err != nil {
-    log.Fatal(err)
-}
-log.Print(res) // do something with the result
-```
-
-Launchpad users can also provide the optional User Agent value describing their Edge User's origination request.
-
-```golang
-params := &models.GetGroupedDailyAggsParams{
-    Locale:     models.US,
-    MarketType: models.Stocks,
-    Date:       models.Date(time.Date(2021, 7, 22, 0, 0, 0, 0, time.Local)),
-}
-
-res, err := c.GetGroupedDailyAggs(context.Background(), params,
-    models.RequiredEdgeHeaders("EDGE_USER_ID", "EDGE_USER_IP_ADDRESS"),
-    models.EdgeUserAgent("EDGE_USER_AGENT"),
-)
-if err != nil {
-    log.Fatal(err)
-}
-log.Print(res) // do something with the result
-```
+Example can be found [here](./rest/example/launchpad).
 
 ## WebSocket Client
 
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ log.Print(res) // do something with the result
 
 #### Launchpad usage
 
-Users of the Launchpad project will need to pass in certain headers in order to make API requests.
+Users of the Launchpad product will need to pass in certain headers in order to make API requests.
 
 ```golang
 params := &models.GetGroupedDailyAggsParams{
@@ -146,7 +146,7 @@ params := &models.GetGroupedDailyAggsParams{
 
 res, err := c.GetGroupedDailyAggs(context.Background(), params,
     models.RequiredEdgeHeaders("EDGE_USER_ID", "EDGE_USER_IP_ADDRESS"),
-    models.EdgeUserAgent("EDGE_USER_USER_AGENT"),
+    models.EdgeUserAgent("EDGE_USER_AGENT"),
 )
 if err != nil {
     log.Fatal(err)
--- a/README.md
+++ b/README.md
@@ -115,6 +115,45 @@ if err != nil {
 log.Print(res) // do something with the result
 ```
 
+#### Launchpad usage
+
+Users of the Launchpad project will need to pass in certain headers in order to make API requests.
+
+```golang
+params := &models.GetGroupedDailyAggsParams{
+    Locale:     models.US,
+    MarketType: models.Stocks,
+    Date:       models.Date(time.Date(2021, 7, 22, 0, 0, 0, 0, time.Local)),
+}
+
+res, err := c.GetGroupedDailyAggs(context.Background(), params,
+    models.RequiredEdgeHeaders("EDGE_USER_ID", "EDGE_USER_IP_ADDRESS"),
+)
+if err != nil {
+    log.Fatal(err)
+}
+log.Print(res) // do something with the result
+```
+
+Launchpad users can also provide the optional User Agent value describing their Edge User's origination request.
+
+```golang
+params := &models.GetGroupedDailyAggsParams{
+    Locale:     models.US,
+    MarketType: models.Stocks,
+    Date:       models.Date(time.Date(2021, 7, 22, 0, 0, 0, 0, time.Local)),
+}
+
+res, err := c.GetGroupedDailyAggs(context.Background(), params,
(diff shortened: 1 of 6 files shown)


jbonzo:
@morningvera I believe this is the approach that makes sense for the library. Let me know if there is a better route.

Also the way it's laid out now will require users to pass all 3 `RequestOption` function but I kind of think I should make a single function that wraps all 3 options in order to enforce that each are passed. I know the APIs will enforce it but this is more of a nice gaurdrail to make it easier to use.

Something like `func RequiredHeaderFnName(edgeUserID, edgeUserIP, edgeUserUserAgent string) RequestOption`

morningvera:
@jbonzo Yeah, this generally looks good to me. I agree with adding that wrapper but `edgeUserAgent` is actually not required by the API. We should make that one optional if it's easy enough to do.

jbonzo:
@morningvera Updated. I don't think I'm ready to take this out of draft yet until I figure out a good way to document and illustrate how to use. I was thinking of either putting an example [here](https://github.com/polygon-io/client-go/tree/master/rest/example), or just adding an example on the top level README.

Opinions?

morningvera:
@jbonzo It might be nice to have both so that users have an easily testable example and a nice explainer section in the readme

rest/models/request.go
```diff
@@ -48,3 +48,30 @@ func QueryParam(key, value string) RequestOption {
 		o.QueryParams.Add(key, value)
 	}
 }
+
+// Headers required to use the Launchpad product.
+const (
+	// HeaderEdgeID is a required Launchpad header. It identifies the Edge User requesting data.
+	HeaderEdgeID = "X-Polygon-Edge-ID"
+	// HeaderEdgeIPAddress is a required Launchpad header. It denotes the originating IP Address of the Edge User requesting data.
+	HeaderEdgeIPAddress = "X-Polygon-Edge-IP-Address "
```
morningvera, line 57: just pointing this out: `"X-Polygon-Edge-IP-Address "` -> `"X-Polygon-Edge-IP-Address"`

rest/example/main.go
```diff
@@ -12,6 +13,7 @@ import (
 func main() {
 	getAllTickersSnapshot()
 	listTrades()
+	getAggregateBarsLaunchpad()
```
jbonzo, line 16: I don't like having this example in this file since it's not the main use case. I was thinking of creating a separate `main.go` for it but wasn't sure where to put it. Could I make a subdir and keep the existing example in `rest/example/common/main.go` and move the new one to `rest/example/launchpad/main.go`?

rest/models/request.go
```diff
@@ -48,3 +48,30 @@ func QueryParam(key, value string) RequestOption {
...
+			o.Headers = make(http.Header)
+		}
+
+		o.Headers.Add(HeaderEdgeID, edgeID)
+		o.Headers.Add(HeaderEdgeIPAddress, edgeIPAddress)
+	}
+}
+
+// EdgeUserAgent sets the Launchpad optional header denoting the Edge User's UserAgent.
+func EdgeUserAgent(userAgent string) RequestOption {
```
jbonzo, line 75: I realized that technically it should be 
```suggestion
// EdgeUserUserAgent sets the Launchpad optional header denoting the Edge User's UserAgent.
func EdgeUserUserAgent(userAgent string) RequestOption {
```

README.md
```diff
@@ -115,6 +115,45 @@ if err != nil {
...
+```golang
+params := &models.GetGroupedDailyAggsParams{
+    Locale:     models.US,
+    MarketType: models.Stocks,
+    Date:       models.Date(time.Date(2021, 7, 22, 0, 0, 0, 0, time.Local)),
+}
+
+res, err := c.GetGroupedDailyAggs(context.Background(), params,
+    models.RequiredEdgeHeaders("EDGE_USER_ID", "EDGE_USER_IP_ADDRESS"),
+    models.EdgeUserAgent("EDGE_USER_USER_AGENT"),
```
morningvera, line 149: make sure to change this as well if you decide to go with `EdgeUserUserAgent`

rest/example/main.go
morningvera: Another option is you could add an arg that let's people choose which example to run, e.g. `POLYGON_API_KEY=... go run rest/example/main.go launchpad`. I'm fine with separate examples too though but we'll need to make sure we update any broken links

rest/models/request.go
jbonzo: Totally agree, but is `EdgeUserAgent` still accurate. If you say yes then I'm convinced, just being a little pedantic.

rest/example/main.go
jbonzo: I like that. I'll make that change

rest/models/request.go
morningvera: From a clarity standpoint, EdgeUserAgent is better imo. It also matches the naming of these other headers and vars in this PR (e.g. you went with HeaderEdgeID instead of HeaderEdgeUserID which I agree with)

morningvera (dismissed):
lgtm

README.md
```diff
@@ -115,6 +115,45 @@ if err != nil {
 log.Print(res) // do something with the result
 ```
 
+#### Launchpad usage
+
+Users of the Launchpad project will need to pass in certain headers in order to make API requests.
```
timetraveler328, line 120: product instead of project?

timetraveler328: "Launchpad product"

timetraveler328:
What if we put the launchpad code in its own file, like rest/example/launchpad/README.md and rest/example/launchpad/main.go?  Then we could link to the github page of rest/example/launchpad from the launchpad docs and all of the relevant info would be there.  And maybe just have a brief mention in the main README with a link to the launchpad README?

jbonzo:
> What if we put the launchpad code in its own file, like rest/example/launchpad/README.md and rest/example/launchpad/main.go? Then we could link to the github page of rest/example/launchpad from the launchpad docs and all of the relevant info would be there. And maybe just have a brief mention in the main README with a link to the launchpad README?

Moved around.

rest/example/main.go
```diff
@@ -2,11 +2,10 @@ package main
 
 import (
 	"context"
-	"log"
-	"os"
-
 	polygon "github.com/polygon-io/client-go/rest"
 	"github.com/polygon-io/client-go/rest/models"
+	"log"
+	"os"
```
morningvera, line 8: Is this an artifact of another change? I recommend reverting it

rest/example/launchpad/README.md
```diff
@@ -0,0 +1,37 @@
+
+Users of the Launchpad product will need to pass in certain headers in order to make API requests.
```
morningvera, line 2: Can you add headers to this? e.g. it should start with something like `# Launchpad`. I kinda disagree with moving this to a separate readme because now it's significantly harder for users to find it.

rest/example/launchpad/main.go
```diff
@@ -0,0 +1,35 @@
+package main
+
+import (
```
morningvera, line 3: It might be worth adding another command to the Makefile for this. We have `make rest-example` and `make ws-example` so maybe something like `make launchpad-example`. Up to you if you want to add it though

rest/example/launchpad/README.md
morningvera: Never mind, I guess the link from the main readme solves that

README.md
```diff
@@ -115,6 +115,11 @@ if err != nil {
 log.Print(res) // do something with the result
 ```
 
+#### Launchpad usage
```
morningvera, line 118: `Launchpad usage` -> `Launchpad Usage` to follow the style of the rest of the repo

rest/example/launchpad/main.go
jbonzo: added

rest/example/launchpad/README.md
jbonzo: Yeah I think you make a decent case. Katie's point of moving it was to make it easier to link from an external site, so if you think that can be solved while keeping them in the same README then please lmk

morningvera: nah, i think this is okay as is but i still recommend adding a header at the top

morningvera (approved):

timetraveler328 (approved):

Issue #210, opened 2022-11-24
Question: Why is Exchange of type *int in ListTickersParams?


ClaytonNorthey92:
Looking at `ListTickersParams`, it is [defined](https://github.com/polygon-io/client-go/blob/c8894b4651c04f28f04d9f60d3d5e8d176638c97/rest/models/tickers.go#L21) as:
```go
// Specify the primary exchange of the asset in the ISO code format. Find more information about the ISO codes at
// the ISO org website. Defaults to empty string which queries all exchanges.
Exchange *int `query:"exchange"`
```

Notice it is of type `*int`, I am wondering why that is?

From the documentation comment above the field:
> ...Defaults to empty _**string**_ which queries all exchanges.

Also, when I use Polygon's website, I enter a string for the `exchange` parameter, for example if I wanted to filter for `XNAS`
```
https://api.polygon.io/v3/reference/tickers?exchange=XNAS&apiKey=mykey
```
<img width="675" alt="image" src="https://user-images.githubusercontent.com/8801919/203809211-17e84f3c-d323-4b1f-8bda-a39d1e73c8ca.png">

This gives me results like so:
```json
{
    "results": [
        {
            "ticker": "AACG",
            "name": "ATA Creativity Global American Depositary Shares",
            "market": "stocks",
            "locale": "us",
            "primary_exchange": "XNAS",
            "type": "ADRC",
            "active": true,
            "currency_name": "usd",
            "cik": "0001420529",
            "composite_figi": "BBG000V2S3P6",
            "share_class_figi": "BBG001T125S9",
            "last_updated_utc": "2022-11-23T00:00:00Z"
        },
        ...
    ]
}
```

shouldn't this be of type `*string` in `ListTickersParams`?

morningvera:
Hey @ClaytonNorthey92, thanks for submitting this! This does seem to be a bug in our client. I'll look into it a bit more and make the fix when I get a chance!

ClaytonNorthey92:
hey @morningvera thank you for the response, I had been working off of a commit in a fork of mine, I just opened a PR for it.  I hope that helps.

PR #218, opened 2022-11-30
Using string instead of int for Exchange in ListTickersParams

ClaytonNorthey92:
fixes https://github.com/polygon-io/client-go/issues/210


morningvera (approved):
lgtm, going to merge and release this shortly. thanks again for the contribution @ClaytonNorthey92!

PR #220, opened 2022-12-05
add summaries endpoint

chaig15:


rest/models/summaries.go
```diff
@@ -0,0 +1,54 @@
...
+	Open                      float64 `json:"open,omitempty"`
+	PreviousClose             float64 `json:"previous_close,omitempty"`
+	Volume                    float64 `json:"volume,omitempty"`
+}
+
+//easyjson:json
+type Option struct {
+	ContractType      string                       `json:"contract_type,omitempty"`
+	ExerciseStyle     string                       `json:"exercise_style,omitempty"`
+	ExpirationDate    ptime.IDaysPolygonDateString `json:"expiration_date,omitempty"`
```
jbonzo, line 50: You don't want to use `ptime` here. Currently this repo has no dependency on `github.com/polygon-io/ptime` even though we have other models that represent time similar to `ptime.IDaysPolygonDateString`.

We don't want to unnecessarily introduce new dependencies so go find another example where we have a field similar to `ptime.IDaysPolygonDateString` and see how this codebase solves it.

jbonzo (approved):

rest/summaries.go
```diff
@@ -0,0 +1,26 @@
...
+const (
+	GetSummariesPath = "/v1/summaries"
+)
+
+// SummariesClient defines a REST client for the Polygon Snapshot Summary API.
+type SummariesClient struct {
+	client.Client
+}
+
+// GetSummaries retrieves summaries for the ticker list with the given params
```
morningvera, line 20: Nit: add a period at the end of this sentence (it'll render to `... given params For more details ...` in the godocs)

morningvera (approved):

PR #222, opened 2022-12-24
Options Chain

antdjohns:
these changes should not be merged until options chain endpoint is complete.

Tests not implemented.

### DIFF
--- a/rest/models/snapshot.go
+++ b/rest/models/snapshot.go
@@ -69,6 +69,115 @@ func (p GetGainersLosersSnapshotParams) WithIncludeOTC(q bool) *GetGainersLosers
 	return &p
 }
 
+// GetOptionsChainParams to get the snapshot of an option contract for a stock equity
+type GetOptionsChainParams struct {
+	// The ticker for the option contract.
+	Ticker *string `validate:"required" path:"underlyingAsset"`
+
+	// The strike price of the option contract.
+	StrikePrice *string `query:"string_price"`
+
+	// The type of contract. Can be "put", "call", or in some rare cases, "other".
+	ContractType *string `query:"contract_type"`
+
+	// The contract's expiration date in YYYY-MM-DD format.
+	ExpirationDateEQ  *string `query:"expiration_date"`
+	ExpirationDateLT  *string `query:"expiration_date.lt"`
+	ExpirationDateLTE *string `query:"expiration_date.lte"`
+	ExpirationDateGT  *string `query:"expiration_date.gt"`
+	ExpirationDateGTE *string `query:"expiration_date.gte"`
+
+	Limit *int   `query:"Limit"`
+	Sort  *Sort  `query:"sort"`
+	Order *Order `query:"order"`
+}
+
+func NewGetOptionsChainParams(ticker string) *GetOptionsChainParams {
+	return &GetOptionsChainParams{Ticker: &ticker}
+}
+
+func (o *GetOptionsChainParams) WithTicker(ticker string) *GetOptionsChainParams {
+	o.Ticker = &ticker
+	return o
+}
+
+func (o *GetOptionsChainParams) WithStrikePrice(strikePrice string) *GetOptionsChainParams {
+	o.StrikePrice = &strikePrice
+	return o
+}
+
+func (o *GetOptionsChainParams) WithContractType(contractType string) *GetOptionsChainParams {
+	o.ContractType = &contractType
+	return o
+}
+
+// WithLimit sets number of results returned. Limit default is 10. Limit must fall in range of 0-1000.
+func (o *GetOptionsChainParams) WithLimit(limit int) *GetOptionsChainParams {
+	o.Limit = &limit
+	return o
+}
+
+// WithExpirationDate sets expiration_date query parameter.
+// comparator Comparator options include EQ, LT, LTE, GT, and GTE.
+// expirationDate should be in YYYY-MM-DD format
+func (o *GetOptionsChainParams) WithExpirationDate(comparator Comparator, expirationDate string) *GetOptionsChainParams {
+	switch comparator {
+	case EQ:
+		o.ExpirationDateEQ = &expirationDate
+	case LT:
+		o.ExpirationDateLT = &expirationDate
+	case LTE:
+		o.ExpirationDateLTE = &expirationDate
+	case GT:
+		o.ExpirationDateGT = &expirationDate
+	case GTE:
+		o.ExpirationDateGTE = &expirationDate
+	default:
+		o.ExpirationDateEQ = &expirationDate
+	}
+	return o
+}
+
+// WithOrder sets order of results based on the Sort field.
+func (o *GetOptionsChainParams) WithOrder(order Order) *GetOptionsChainParams {
+	switch order {
+	case Asc:
+		o.Order = &order
+	case Desc:
+		o.Order = &order
+	}
+	return o
+}
+
+// WithSort sets sort field. Sort expects to receive TickerSymbol, ExpirationDate, or StrikePrice as an argument.
+func (o *GetOptionsChainParams) WithSort(sort Sort) *GetOptionsChainParams {
+	switch sort {
+	case TickerSymbol:
+		o.Sort = &sort
+	case ExpirationDate:
+		o.Sort = &sort
+	case StrikePrice:
+		o.Sort = &sort
+	}
+	return o
+}
+
+type GetOptionsChainSnapshotResponse struct {
+	BaseResponse
+	Results []OptionsChain `json:"results,omitempty"`
+}
+
+type OptionsChain struct {
+	Day               DaySnapshot     `json:"day,omitempty"`
+	Details           OptionDetails   `json:"details,omitempty"`
+	Greeks            Greeks          `json:"greeks,omitempty"`
+	LastQuote         LastQuote       `json:"last_quote,omitempty"`
+	UnderlyingAsset   UnderlyingAsset `json:"underlying_asset,omitempty"`
+	ImpliedVolatility *float64        `json:"implied_volatility,omitempty"`
+	OpenInterest      *float64        `json:"open_interest,omitempty"`
+	BreakEvenPrice    *float64        `json:"break_even_price,omitempty"`
+}
+
 // GetGainersLosersSnapshotResponse is the response returned by the GetGainersLosersSnapshot method.
 type GetGainersLosersSnapshotResponse struct {
 	BaseResponse
--- a/rest/models/snapshot.go
+++ b/rest/models/snapshot.go
@@ -71,7 +71,7 @@ func (p GetGainersLosersSnapshotParams) WithIncludeOTC(q bool) *GetGainersLosers
 
 // ListOptionsChainParams is a set of parameters for the ListOptionsChainSnapshot method.
 type ListOptionsChainParams struct {
-	// The ticker for the option contract.
+	// The underlying ticker symbol of the option contract.
 	Ticker string `validate:"required" path:"underlyingAsset"`
 
 	// The strike price of the option contract.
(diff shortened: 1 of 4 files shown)


rest/models/snapshot.go
```diff
@@ -69,6 +69,115 @@ func (p GetGainersLosersSnapshotParams) WithIncludeOTC(q bool) *GetGainersLosers
...
+}
+
+// WithLimit sets number of results returned. Limit default is 10. Limit must fall in range of 0-1000.
+func (o *GetOptionsChainParams) WithLimit(limit int) *GetOptionsChainParams {
+	o.Limit = &limit
+	return o
+}
+
+// WithExpirationDate sets expiration_date query parameter.
+// comparator Comparator options include EQ, LT, LTE, GT, and GTE.
```
antdjohns, line 121:
```suggestion
// comparator options include EQ, LT, LTE, GT, and GTE.
```

rest/models/snapshot.go
```diff
@@ -69,6 +69,115 @@ func (p GetGainersLosersSnapshotParams) WithIncludeOTC(q bool) *GetGainersLosers
 	return &p
 }
 
+// GetOptionsChainParams to get the snapshot of an option contract for a stock equity
```
antdjohns, line 72:
```suggestion
// GetOptionsChainParams to get the snapshot of all options contracts for an underlying ticker.
```

antdjohns, line 72:
```suggestion
// GetOptionsChainParams is a set of parameters for the GetOptionsChainSnapshot method.
```

rest/models/snapshot.go
```diff
@@ -69,6 +69,120 @@ func (p GetGainersLosersSnapshotParams) WithIncludeOTC(q bool) *GetGainersLosers
...
+	Order *Order `query:"order"`
+}
+
+// NewGetOptionsChainParams will return a pointer to GetOptionsChainParams.
+func NewGetOptionsChainParams(ticker string) *GetOptionsChainParams {
+	return &GetOptionsChainParams{Ticker: &ticker}
+}
+
+// WithTicker sets underlying ticker symbol to params
+func (o *GetOptionsChainParams) WithTicker(ticker string) *GetOptionsChainParams {
```
jbonzo, line 101: Other endpoints in this repo don't have the method receiver as a pointer.

rest/models/snapshot.go
```diff
@@ -69,6 +69,120 @@ func (p GetGainersLosersSnapshotParams) WithIncludeOTC(q bool) *GetGainersLosers
...
+}
+
+// WithOrder sets order of results based on the Sort field.
+func (o *GetOptionsChainParams) WithOrder(order Order) *GetOptionsChainParams {
+	switch order {
+	case Asc:
+		o.Order = &order
+	case Desc:
+		o.Order = &order
+	}
```
jbonzo, line 153: No need for a switch case

jbonzo (commented):
Overall comment is to follow the endpoints to a T. This repo should be uniform and clean

rest/models/snapshot.go
```diff
@@ -69,6 +69,120 @@ func (p GetGainersLosersSnapshotParams) WithIncludeOTC(q bool) *GetGainersLosers
...
+	ContractType *ContractType `query:"contract_type"`
+
+	// The contract's expiration date in YYYY-MM-DD format.
+	ExpirationDateEQ  *string `query:"expiration_date"`
+	ExpirationDateLT  *string `query:"expiration_date.lt"`
+	ExpirationDateLTE *string `query:"expiration_date.lte"`
+	ExpirationDateGT  *string `query:"expiration_date.gt"`
+	ExpirationDateGTE *string `query:"expiration_date.gte"`
+
+	Limit *int   `query:"Limit"`
```
morningvera, line 90: This should be `query:"limit"`

rest/models/snapshot.go
```diff
@@ -69,6 +69,120 @@ func (p GetGainersLosersSnapshotParams) WithIncludeOTC(q bool) *GetGainersLosers
...
+	Results []OptionsChainSnapshot `json:"results,omitempty"`
+}
+
+type OptionsChainSnapshot struct {
+	Day               DaySnapshot     `json:"day,omitempty"`
+	Details           OptionDetails   `json:"details,omitempty"`
+	Greeks            Greeks          `json:"greeks,omitempty"`
+	LastQuote         LastQuote       `json:"last_quote,omitempty"`
+	UnderlyingAsset   UnderlyingAsset `json:"underlying_asset,omitempty"`
+	ImpliedVolatility *float64        `json:"implied_volatility,omitempty"`
```
morningvera, line 181: Response fields should never be pointers

rest/models/types.go
```diff
@@ -15,6 +15,15 @@ const (
 	Crypto MarketType = "crypto"
 )
 
+// ContractType is the type of contract.
+type ContractType string
+
+const (
+	ContractTypeCall  ContractType = "call"
```
morningvera, line 22: I think `ContractCall` would more closely match the style of the other enums in this repo

rest/snapshot.go
```diff
@@ -69,6 +70,14 @@ func (ac *SnapshotClient) GetOptionContractSnapshot(ctx context.Context, params
 	return res, err
 }
 
+// GetOptionsChainSnapshot gets the snapshot of all options contracts for an underlying ticker. For more details see
+// https://polygon.io/docs/options/get_v3_snapshot_options__underlyingasset
+func (ac *SnapshotClient) GetOptionsChainSnapshot(ctx context.Context, params *models.GetOptionsChainParams, opts ...models.RequestOption) (*models.GetOptionsChainSnapshotResponse, error) {
```
morningvera, line 75: AFAIK this method should actually be `ListOptionsChainSnapshot` because the response is paginated and includes a `next_url`. It should return an iterator similar to our other list methods. If that's a correct assumption on my part, can we also rename the relevant types to `List*`?

rest/models/snapshot.go
```diff
@@ -69,6 +69,120 @@ func (p GetGainersLosersSnapshotParams) WithIncludeOTC(q bool) *GetGainersLosers
...
+	}
+	return o
+}
+
+type GetOptionsChainSnapshotResponse struct {
+	BaseResponse
+	Results []OptionsChainSnapshot `json:"results,omitempty"`
+}
+
+type OptionsChainSnapshot struct {
```
morningvera, line 175: Could this type be a duplicate of `OptionsContractSnapshot`? I see a couple minor differences (e.g. this one uses `DaySnapshot` instead of `DayOptionContractSnapshot`) so I could be wrong. If it makes sense though, I suggest deleting this and changing the above results field to `Results []OptionsContractSnapshot`

morningvera: I just checked the docs on staging and it does look like reusing `OptionsContractSnapshot` here is the right choice

rest/models/snapshot.go
```diff
@@ -69,6 +69,120 @@ func (p GetGainersLosersSnapshotParams) WithIncludeOTC(q bool) *GetGainersLosers
 	return &p
 }
 
+// GetOptionsChainParams is a set of parameters for the GetOptionsChainSnapshot method.
+type GetOptionsChainParams struct {
+	// The ticker for the option contract.
+	Ticker *string `validate:"required" path:"underlyingAsset"`
```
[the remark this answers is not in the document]
antdjohns: How do we determine what should be a pointer and what should not? anything in the path is not a pointer while query values are pointers?
That seems to be the patten, just clarifying.

rest/models/snapshot.go
```diff
@@ -69,6 +69,120 @@ func (p GetGainersLosersSnapshotParams) WithIncludeOTC(q bool) *GetGainersLosers
...
+
+	// The strike price of the option contract.
+	StrikePrice *string `query:"string_price"`
```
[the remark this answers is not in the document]
antdjohns: Changed to *float64.
What is RHS?

rest/models/snapshot.go
```diff
@@ -69,6 +69,120 @@ func (p GetGainersLosersSnapshotParams) WithIncludeOTC(q bool) *GetGainersLosers
...
+	ExpirationDateGT  *string `query:"expiration_date.gt"`
+	ExpirationDateGTE *string `query:"expiration_date.gte"`
+
+	Limit *int   `query:"Limit"`
+	Sort  *Sort  `query:"sort"`
+	Order *Order `query:"order"`
+}
+
+// NewGetOptionsChainParams will return a pointer to GetOptionsChainParams.
+func NewGetOptionsChainParams(ticker string) *GetOptionsChainParams {
```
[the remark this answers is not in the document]
antdjohns: Removed

rest/models/snapshot.go
antdjohns: changed pointer receiver to non-pointer.

rest/models/snapshot.go
antdjohns: done

rest/models/snapshot.go
antdjohns: Removed this struct and replaced with the OptionsContractSnapshot struct that was already being used.

rest/models/types.go
antdjohns: Changed

rest/snapshot.go
antdjohns: Made changes, good callout.
changed response to *iter.Iter[models.OptionsContractSnapshot] and renamed appropriately

rest/models/snapshot.go
antdjohns: Just now seeing this. and yes, I made the change.

rest/models/snapshot.go
```diff
@@ -69,6 +69,104 @@ func (p GetGainersLosersSnapshotParams) WithIncludeOTC(q bool) *GetGainersLosers
 	return &p
 }
 
+// ListOptionsChainParams is a set of parameters for the ListOptionsChainSnapshot method.
+type ListOptionsChainParams struct {
+	// The underlying ticker symbol of the option contract.
+	UnderlyingAsset string `validate:"required" path:"underlyingAsset"`
+
+	// The strike price of the option contract.
+	StrikePrice *float64 `query:"string_price"`
```
jbonzo, line 78:
```suggestion
	StrikePrice *float64 `query:"strike_price"`
```

rest/models/snapshot.go
```diff
@@ -69,6 +69,104 @@ func (p GetGainersLosersSnapshotParams) WithIncludeOTC(q bool) *GetGainersLosers
...
+
+	// Sort field used for ordering.
+	Sort *Sort `query:"sort"`
+
+	// Order results based on the sort field.
+	Order *Order `query:"order"`
+}
+
+// WithUnderlyingAsset the underlying ticker symbol of the option contract.
+func (o ListOptionsChainParams) WithUnderlyingAsset(ticker string) *ListOptionsChainParams {
```
morningvera, line 101: We generally only provide these helpers for query params since they're pointers and a bit annoying to use without helpers. The usage pattern is something like this:
```golang
res, err := c.GetDailyOpenCloseAgg(context.Background(), models.GetDailyOpenCloseAggParams{
	Ticker: "AAPL",
	Date:   models.Date(time.Date(2020, 10, 14, 0, 0, 0, 0, time.Local)),
}.WithAdjusted(true))
```

rest/models/snapshot.go
```diff
@@ -132,7 +230,7 @@ type LastQuoteSnapshot struct {
 	Timestamp Nanos   `json:"t,omitempty"`
 }
 
-// LastQuoteSnapshot is the most recent trade for a ticker.
+// LastTradeSnapshot is the most recent trade for a ticker.
```
morningvera, line 233: Thanks for these doc fixes!

morningvera (approved):
one more minor thing but otherwise lgtm

rest/models/snapshot.go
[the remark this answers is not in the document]
jbonzo: "right hand side"

rest/models/snapshot.go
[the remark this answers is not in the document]
jbonzo: That is the pattern. Required is not a pointer, optional is a pointer.

jbonzo:
Tests are implemented now right?

antdjohns:
> Tests are implemented now right?

yes

rest/models/snapshot.go
antdjohns: makes sense. removing now.

rest/models/snapshot.go
antdjohns: np

jbonzo (approved):

jbonzo:
We are good to merge

PR #227, opened 2023-01-09
Implement Ticker Events

jbonzo:


rest/vx.go
```diff
@@ -39,3 +40,22 @@ func (c *VXClient) ListStockFinancials(ctx context.Context, params *models.ListS
 		return res, res.Results, err
 	})
 }
+
+// GetTickerEvents retrieves a timeline of events for the entity associated with the given ticker, CUSIP, or Composite FIGI.
+//
+// Note: this method utilizes an experimental API and could experience breaking changes or deprecation.
+//
+// This method returns an iterator that should be used to access the results via this pattern:
```
morningvera, line 48: This example isn't relevant here because this isn't a list endpoint

rest/models/tickers.go
```diff
@@ -347,3 +349,46 @@ type TickerType struct {
...
+	// Name is the company name.
+	Name   string        `json:"name"`
+	Events []TickerEvent `json:"events"`
+}
+
+// TickerEvent contains the data for the different type of ticker events that could occur.
+type TickerEvent struct {
+	Date         string             `json:"date"`
+	Type         string             `json:"type"`
+	TickerChange *TickerChangeEvent `json:"ticker_change,omitempty"`
```
morningvera, line 388: Does this need to be a pointer? It's fine if so but afaik it would be the only case where we use pointers in response types

jbonzo: Oh interesting. Well for now I can leave it out. But there is a future where `ticker_change` isn't present in the response.

morningvera (approved):
lgtm

PR #229, opened 2023-01-11
Financials param update

antdjohns:
Addressing [ticket](https://app.zenhub.com/workspaces/backend-6042935fc77665000f50780e/issues/gh/polygon-io/backend/794)

Also ran `make fmt` and `make rest-spec`

### DIFF
--- a/.polygon/rest.json
+++ b/.polygon/rest.json
@@ -445,6 +445,13 @@
               "type": "string"
             }
           },
+          "required": [
+            "id",
+            "type",
+            "market",
+            "name",
+            "url"
+          ],
           "type": "object"
         },
         "type": "array"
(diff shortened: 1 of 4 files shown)


jbonzo (approved):
I'm not sure if we want to run the rest spec update in this PR. I forget if we want that to be its own PR or not. @morningvera ?

morningvera:
@jbonzo It should generally be paired with the PR that's introducing the changes (easier to review/verify that way). It really depends though, sometimes the new spec includes changes to multiple endpoints and might make sense in a separate PR in that case.

morningvera (approved):

jbonzo:
> @jbonzo It should generally be paired with the PR that's introducing the changes (easier to review/verify that way). It really depends though, sometimes the new spec includes changes to multiple endpoints and might make sense in a separate PR in that case.

Gotcha, so we should have updated the spec back when we implemented Options Chain.

Issue #224, opened 2023-01-06
Unmarshal error when running the example code (GetAllTickersSnapshot)


nlalor:
**Describe the bug**

`GetAllTickersSnapshot` fails to unmarshal the `AccumulatedVolume` value on `MinuteSnapshot`.

`2023/01/06 09:29:47.837606 ERROR RESTY json: cannot unmarshal number 1.012021e+06 into Go struct field MinuteSnapshot.tickers.min.av of type int, Attempt 1
2023/01/06 09:29:47 failed to execute request: json: cannot unmarshal number 1.012021e+06 into Go struct field MinuteSnapshot.tickers.min.av of type int`

**To Reproduce**

Run the example code shown in the "Using the Client" section of the readme. No changes to the example code are needed to reproduce the error on my end.

**Expected behavior**

`GetAllTickersSnapshot` should correctly handle the accumulated volume data.

**Screenshots**

**Additional context**

Go version 1.19

I was able to "fix" this error locally by changing the data type for the `AccumulatedValue` field to `float64` instead of `int`.
Not sure if the schema is just out of date or if there may be some versioning issue.

antdjohns:
Hey @nlalor, We were able to reproduce the issue. Thank you for bringing it to our attention.

morningvera:
This was closed by #230. Thanks for the bug report @nlalor and for the fix @antdjohns!

PR #230, opened 2023-01-20
Accumulated Volume wrong type causing unmarshalling errors in the client

antdjohns:
Addressing GH [ticket](https://github.com/polygon-io/client-go/issues/224)

MinuteSnapshot is using the wrong type for the AccumulatedVolume field. A user reported an error with unmarshalling that I was able to reproduce.


nathanhack:
This would be nice to have.

nathanhack:
That said, the api from polygon.io has the field type set as a `number` (instead of an `integer`) which would imply there exist cases where the volume could be fractional. Is that possible?

antdjohns:
> That said, the api from polygon.io has the field type set as a `number` (instead of an `integer`) which would imply there exist cases where the volume could be fractional. Is that possible?

Yes, stock splits and acquisitions can cause fractional shares.

nathanhack:
> 

Oh good point, I didn't think of either acquisitions or reducing splits. I stand by my, `good to have` position.

morningvera (approved):
@antdjohns lgtm but can you also change the test? changing this to a float is probably sufficient: https://github.com/polygon-io/client-go/blob/master/rest/snapshot_test.go#L41

antdjohns:
@morningvera @jbonzo updated the test data to use a float; would you like me to merge? and if releasing can I include the version patch update in this pr or cut another pr?

morningvera:
@antdjohns yeah, you can merge it but let’s wait on releasing until tomorrow

PR #251, opened 2023-02-28
Add support for indices

jbonzo:


jbonzo:
@morningvera This is my first time updating the websocket side here. Please lmk if I'm missing something when it comes to the new event type `V`

qrpike (approved):

websocket/models/models.go
```diff
@@ -360,3 +360,18 @@ type Level2Book struct {
 	// The timestamp that the tick was received by Polygon.
 	ReceivedTimestamp int64 `json:"r,omitempty"`
 }
+
+// IndexValue is value data for either indices.
+type IndexValue struct {
```
morningvera, line 365: Is this the name we're using in the API docs too? I don't love adding `Value` at the end of types because it should be self explanatory if a descriptive enough name is chosen

morningvera (approved):
one minor thing but otherwise lgtm!

websocket/models/models.go
jbonzo: `value` is the data type. It is the same class as `trade`, `price`, or `quote`

aitzkovitz:
I think you'll also have to add a `Value` field to the UnderlyingAsset model in rest/models/snapshot.go to support the underlying asset for indices contracts

antdjohns (approved):
changes for market status look correct.

aitzkovitz (approved):

jbonzo:
Requesting re-reviews since I added index snapshot. Nothing changed besides the content of [this commit](https://github.com/polygon-io/client-go/pull/251/commits/9b9b1ab5ae45e552b08b77b62ab2f0890b679e63)

morningvera (approved):

PR #259, opened 2023-03-13
Update minor version to v1.11.0

jbonzo:


jbonzo:
Apparently we were already at `v1.10`, need to update this PR

morningvera (approved):
good catch, lgtm

PR #265, opened 2023-03-24
Exclude example files from coverage report

timetraveler328:


.github/workflows/test-coverage.yml
```diff
@@ -21,7 +21,9 @@ jobs:
           fetch-depth: 0 # otherwise, there would be errors pushing refs to the destination repository.
       - name: go-test
         run: |
-          go test -v ./... -covermode=count -coverprofile=coverage.out
+          go test -v ./... -covermode=count -coverprofile=coverage.out.temp
+          cat coverage.out.temp | grep -v "/example" > coverage.out
```
joedursun, line 25: Looks good! I think we can remove the `cat` here and just modify the grep to:

```suggestion
          grep -v "/example" coverage.out.temp > coverage.out
```

jbonzo (approved):

PR #266, opened 2023-03-27
enhance websocket support for indices

moeshinyo:


jbonzo (approved):
Thanks for contributing!

PR #270, opened 2023-04-06
rest/models: print out timestamps in more appealing format

kevinburke:
Previously if you printed a models.Millis with `%#v` you would get something that looked like this:

    models.Millis{wall:0x0, ext:63816264000, loc:(*time.Location)(0x105256980)

Which doesn't tell you at a glance what time it represents. This change makes it print out a prettier looking string that tells you what time it represents.

    Millis(time.Date(2023, time.April, 3, 21, 0, 0, 0, time.Local)

I previously merged the same change to the Go standard library.


justinpolygon:
Hey @kevinburke, thanks for submitting this. I wanted to learn about your use-case for this a little. We have been exploring this idea for our client libraries and wanted to add helper functions that turn unix time into something human readable (1692115582246 -> 2023-08-15 12:06:22.246135-04:00). I just wanted to learn about your use-case for why you'd need it in this format `time.Date(2023, time.April, 3, 21, 0, 0, 0, time.Local)` vs  something like 2023-08-15 12:06:22.246135-04:00. I'm assuming it's so that you can just copy/paste into areas of your code but wanted to check.

kevinburke:
apologies - I can't even remember the context where I encountered this library, sorry

justinpolygon:
You're good @kevinburke. Was just looking at something related and was going to merge this. Also, it looks like we can use something like this for `Nanos` as well. Cheers. Sorry for the massive delay here.

justinpolygon:
Going to refactor a big and submit my own PR. I'll close this out for now.

PR #272, opened 2023-04-13
Fix crypto snapshot unmarshal

justinpolygon:
If you try and fetch crypto snapshots with GetTickerSnapshot, GetAllTickersSnapshot, or GetGainersLosersSnapshot you'll get an error saying that the response `cannot unmarshal number 0.25 into Go struct field LastTradeSnapshot.ticker.lastTrade.s of type int`. The reason is that [Crypto Snapshots - All Tickers](https://polygon.io/docs/crypto/get_v2_snapshot_locale_global_markets_crypto_tickers) and other Crypto Snapshots can return floats vs ints for partial shares.

The fix it to update `LastTradeSnapshot`'s `Size` to be `float64` vs `int`.

Here's some test code:

```golang
package main

import (
	"context"
	"log"
	"os"

)

func main() {
	// init client
	c := polygon.New(os.Getenv("POLYGON_API_KEY"))

	// set params
	params := &models.GetAllTickersSnapshotParams{
		Locale:     models.Global,
		MarketType: models.Crypto,
	}

	// make request
	res, err := c.GetAllTickersSnapshot(context.Background(), params)
	if err != nil {
		log.Fatal(err)
	}

	// do something with the result
	log.Print(res) 
}
```

Here's the error message:
```
2023/04/13 07:21:22.743432 ERROR RESTY json: cannot unmarshal number 206.8 into Go struct field LastTradeSnapshot.tickers.lastTrade.s of type int, Attempt 1
2023/04/13 07:21:22 failed to execute request: json: cannot unmarshal number 206.8 into Go struct field LastTradeSnapshot.tickers.lastTrade.s of type int
exit status 1
```

Tested this on other snapshot examples for stocks, options, indices, and forex with nothing seeming to not break.


jbonzo (approved):
This is a breaking change. This is fine since its fixing a bug, but @justinpolygon can you please consider this change in the release notes as breaking. See [this as an example](https://github.com/polygon-io/client-go/releases/tag/v1.10.0)

justinpolygon:
Thanks @jbonzo. Will do re: release notes.

PR #273, opened 2023-04-13
Add crypto, forex, and indices examples

justinpolygon:
Add examples for Crypto, Forex, and Indices.


jbonzo (approved):

morningvera (approved):
This is a lot of directories! Maybe it'll be cleaner if we organize things like this:

```
├── rest
│   ├── example
│   │   ├── crypto
|   │   │   ├── aggregates-bars
|   |   │   │   ├── main.go
|   │   │   ├── conditions
|   |   │   │   ├── main.go
│   │   ├── indices
|   │   │   ├── aggregates-bars
|   |   │   │   ├── main.go
```

Also, have you considered using [Go examples](https://go.dev/blog/examples)? It might not make sense here but it is very nice that the examples show up in the godoc

justinpolygon:
Hey @morningvera, this is a great idea around the directory structure. I totally agree it's a ton of directories. I'll refactor on my end here. Also, I haven't seen that before where the examples show up in the godoc. I'll explore that too.

justinpolygon:
FYI - @morningvera moved things around liked we chatted about. Thanks for the great suggestion. I'll merge now.

PR #279, opened 2023-04-28
move rawData output into processData()

istrau2:
The client does a bunch of processing prior to returning raw data which can impact performance.

In my case, I would like to:
- reduce number of json deserializations to 1 (currently, even if consuming raw data, one must do 3 since 2 calls to json.Unmarshall() prior to pushing out the raw data)
- moving to a different more performant json deserialization package (many to choose from)

### DIFF
--- a/websocket/polygon.go
+++ b/websocket/polygon.go
@@ -43,9 +43,10 @@ type Client struct {
 	wQueue chan json.RawMessage
 	subs   subscriptions
 
-	rawData bool
-	output  chan any
-	err     chan error
+	authenticated bool
+	rawData       bool
+	output        chan any
+	err           chan error
 
 	log Logger
 }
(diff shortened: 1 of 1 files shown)


websocket/polygon.go
```diff
@@ -313,6 +314,7 @@ func (c *Client) write() error {
 	defer func() {
 		c.log.Debugf("write thread closed")
 		ticker.Stop()
+		c.authenticated = false
```
jbonzo, line 317: Can you explain the changes to the `authenticated` flag? I don't see its relevance to processing rawData.

jbonzo: Nevermind I understand its functionality now.

jbonzo (commented):
I understand the goal if this PR but I am a little weary of the incorporation of the additional authentication logic.

My understanding of this addition is that you, the user, want to handle all data type events but have the client handle all other types. I made a draft PR #305 that proposes an alternative approach where the user handles all event types returned by the API. The idea is to keep the functionality simple and the contract of this feature small and predictable.

Let me know your thoughts and if #305 satisfies your needs.

jbonzo:
Closing in favor of #302

PR #283, opened 2023-05-09
Luke/backend/1036/snapshot v3 endpoint

lukeoleson:


rest/example/options/snapshots-assets/main.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/polygon-io/client-go/rest/models"
+)
+
+func main() {
+	// Init client
+	c := polygon.New(os.Getenv("POLYGON_API_KEY"))
+
+	// Set parameters
+	params := models.ListAssetSnapshotsParams{}.
+		WithTickerAnyOf("AAPL,META,F")
```
aitzkovitz, line 18: since this example is in the options directory, probably makes more sense to make these options tickers

lukeoleson: Ah, good catch! Thanks :)

rest/models/snapshot.go
```diff
@@ -378,3 +380,97 @@ type OrderBookQuote struct {
...
+	Name         string            `json:"name,omitempty"`
+	MarketStatus string            `json:"market_status,omitempty"`
+	Ticker       string            `json:"ticker,omitempty"`
+	Type         string            `json:"type,omitempty"`
+	LastQuote    SnapshotLastQuote `json:"last_quote,omitempty"`
+	LastTrade    SnapshotLastTrade `json:"last_trade,omitempty"`
+	Session      Session           `json:"session,omitempty"`
+
+	BreakEvenPrice      float64         `json:"break_even_price,omitempty"`
+	Details             DetailsV2       `json:"details,omitempty"`
```
aitzkovitz, line 440: We probably don't need the V2 on this, unless there's another type in this package with that name.

rest/models/snapshot.go
```diff
@@ -378,3 +380,97 @@ type OrderBookQuote struct {
...
+	}
+	return &p
+}
+
+type ListAssetSnapshotsResponse struct {
+	BaseResponse
+	Results []SnapshotResponseModel `json:"results,omitempty"`
+}
+
+type SnapshotResponseModel struct {
```
aitzkovitz, line 430: These may have mistakenly been left out of the response model which the API defines, but each result can also be an error object and look like this: 
```
    {
      "error": "NOT_FOUND",
      "message": "Ticker not found.",
      "ticker": "TSLAAPL"
    }
```
So we'll probably want to add the `error` and `message` fields to this object.

rest/models/snapshot.go
```diff
@@ -378,3 +380,97 @@ type OrderBookQuote struct {
...
+	Type         string            `json:"type,omitempty"`
+	LastQuote    SnapshotLastQuote `json:"last_quote,omitempty"`
+	LastTrade    SnapshotLastTrade `json:"last_trade,omitempty"`
+	Session      Session           `json:"session,omitempty"`
+
+	BreakEvenPrice      float64         `json:"break_even_price,omitempty"`
+	Details             DetailsV2       `json:"details,omitempty"`
+	Greeks              Greeks          `json:"greeks,omitempty"`
+	ImpliedVolatility   float64         `json:"implied_volatility,omitempty"`
+	Sofr                float64         `json:"sofr,omitempty"`
```
aitzkovitz, line 443: we actually don't ever return `SOFR` or `AnnualDividendsPaid` so we can remove those from this type. They also shouldn't exist on that response model in the API. they only exist on the options snapshot object to be used for other calculations.

rest/models/summaries.go
```diff
@@ -44,6 +44,9 @@ type Session struct {
 	Open                      float64 `json:"open,omitempty"`
 	PreviousClose             float64 `json:"previous_close,omitempty"`
 	Volume                    float64 `json:"volume,omitempty"`
+	EarlyTradingPrice         float64 `json:"early_trading_price,omitempty"`
```
aitzkovitz, line 47: I don't think we ever return these fields from the summaries endpoint, so we shouldn't need them, unless you've seen otherwise

rest/snapshot_test.go
```diff
@@ -596,3 +598,169 @@ func TestGetIndicesSnapshot(t *testing.T) {
...
+		assert.Equal(t, wantSnapshot, iter.Item())
+		iterCount++
+	}
+
+	assert.Equal(t, len(assetSnapshotsTestData), iterCount, fmt.Sprintf("expected %d iterators", len(assetSnapshotsTestData)))
+	assert.False(t, iter.Next())
+	assert.Nil(t, iter.Err())
+}
+
+var assetSnapshotsTestData = []string{
```
aitzkovitz, line 648: Since this endpoint is polymorphic, it might make sense to add different types of snapshots to the response test data. right now, it only has stocks snapshots, but you could add an options snapshot, and and error result to make sure those are unmarshaled correctled

go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/go-resty/resty/v2 v2.7.0
 	github.com/gorilla/websocket v1.5.0
 	github.com/jarcoal/httpmock v1.3.0
+	github.com/polygon-io/ptime v1.3.0
```
aitzkovitz, line 12: i don't think you need this, i don't see it being used anywhere, and its a private library

rest/snapshot_test.go
```diff
@@ -764,3 +817,156 @@ var assetSnapshotsTestData = []string{
...
+		},
+		"greeks": {},
+		"underlying_asset": {
+		  "change_to_break_even": -0.11,
+		  "last_updated": 1683732072879546553,
+		  "price": 173.135,
+		  "ticker": "AAPL",
+		  "timeframe": "REAL-TIME"
+		},
+		"error": "",
```
aitzkovitz, line 854: you can leave these properties off for the successful responses. they'll only be there for error responses

rest/snapshot_test.go
```diff
@@ -605,47 +605,100 @@ func TestListAssetSnapshots(t *testing.T) {
...
+				var gotSnapshot models.SnapshotResponseModel
+				err := json.Unmarshal([]byte(tc.testData[iterCount]), &gotSnapshot)
+				require.Nil(t, err)
+
+				require.Nil(t, iter.Err())
+				assert.Equal(t, gotSnapshot, iter.Item())
+				iterCount++
+			}
+
+			assert.Equal(t, len(tc.testData), iterCount, fmt.Sprintf("expected %d iterators", len(tc.testData)))
```
aitzkovitz, line 694: might make more sense to say "exected %d results"

aitzkovitz (approved):
looks good! just had a few comments

jbonzo (approved):
I tested it locally, works well

PR #285, opened 2023-05-11
Support for go 1.20

jbonzo:


.github/workflows/lint.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        go-version: [1.19.x]
+        go-version: [1.19.x, 1.20.x]
```
jbonzo, line 16: @morningvera do I want both versions or just latest?

morningvera: I think it’s safe to just do latest

morningvera (approved):

PR #290, opened 2023-05-18
Added examples and updated models for forex and crypto for new v3 snapshot endpoint

lukeoleson:
- Added examples and tests for Forex and Crypto Snapshots.
- Added a couple more data points to the response that were missing compared to the return values for crypto/forex requests (`lastTrade.ParticipateTimestamp` and `session.Price`).


rest/snapshot.go
```diff
@@ -17,7 +17,7 @@ const (
 	ListOptionsChainSnapshotPath  = "/v3/snapshot/options/{underlyingAsset}"
 	GetCryptoFullBookSnapshotPath = "/v2/snapshot/locale/global/markets/crypto/tickers/{ticker}/book"
 	GetIndicesSnapshotPath        = "/v3/snapshot/indices"
-	ListAssetSnapshots            = "/v3/snapshot"
+	ListAssetSnapshotsPath        = "/v3/snapshot"
```
jbonzo, line 20: This is a breaking change. Its an exported field that you are removing. Might not matter as much since I haven't merged [my pr](https://github.com/polygon-io/client-go/pull/286/files#diff-27688c042992efb8fb7f4c17257e043c3301e235060ed5ff018437cd4b45dbaaR23) yet which introduces a new name.

jbonzo (commented):
Can you change the title of the PR to be sentence structure and remove your branch name. It'll become the commit name.

rest/example/crypto/snapshots-assets/main.go
jbonzo, line 1: Depending on whether this PR or [this PR](https://github.com/polygon-io/client-go/pull/286/files#diff-63d0f5aa77327ec72e51fdd874e8ce831dbfdea7262aa8fb5e41a46b34b656e7) gets merged first, this filename should change to match.

rest/snapshot.go
lukeoleson: Great catch, thanks!

This is updated to the latest from master (your change in [this pr](https://github.com/polygon-io/client-go/pull/286/files)).

rest/example/crypto/snapshots-assets/main.go
lukeoleson: That one got in first, so I've resolved the conflicts and updated both `rest/example/crypto/snapshots-assets` and `rest/example/forex/snapshots-assets` to use `snapshots-universal` instead.

aitzkovitz (approved):

PR #302, opened 2023-06-12
Allow rawData to bypass any extra JSON processing

jbonzo:


websocket/config.go
```diff
@@ -20,9 +20,16 @@ type Config struct {
 	// client to reconnect indefinitely until the user closes it.
 	MaxRetries *uint64
 
-	// RawData is a flag indicating whether data should be returned as a raw JSON.
+	// RawData is a flag indicating whether data should be returned as a raw JSON or raw bytes. If BypassRawDataRouting is unset
+	// then the data will be returned as raw JSON, otherwise it will be raw bytes.
 	RawData bool
 
+	// BypassRawDataRouting is a flag that interacts with the RawData flag. If RawData flag is unset then this flag is ignore.
```
mmoghaddam385, line 27:
```suggestion
	// BypassRawDataRouting is a flag that interacts with the RawData flag. If RawData flag is unset then this flag is ignored.
```

websocket/config.go
```diff
@@ -20,9 +20,16 @@ type Config struct {
...
+	// If both this flag and RawData is set then this flag indicates whether the raw data should be parsed as raw JSON
+	// and routed via the client's internal logic.
+	// Enabling this flag means it us up to caller to handle all message types including auth and subscription responses.
```
mmoghaddam385, line 30: Some suggestions to make docs for this flag a little more clear, lmk what you think

```suggestion
		// If RawData is `true`, then this flag indicates whether the raw data should be parsed as json.RawMessage
	// and routed via the client's internal logic (`BypassRawDataRouting=false`), or returned to the application code as []byte (`BypassRawDataRouting=true`).
	// If this flag is `true`, it's up to the caller to handle all message types including auth and subscription responses.
```

Issue #192, opened 2022-10-01
failed to unmarshal MinuteSnapshot accumulated volume


AnteWall:
**Describe the bug**
Sometimes when requesting Snapshot.GetTickerSnapshot we get the following error trying to marshal the MinuteSnapshot struct.
`json: cannot unmarshal number 4.578654e+06 into Go struct field MinuteSnapshot.ticker.min.av of type int`

**To Reproduce**
Request GetTickerSnapshot ticker with data that as av that is not a int
I requested ticker `ET` during this weekend (2022-10-01)

morningvera:
Hey @AnteWall, thanks for the bug report! Based on the API spec, that field should always be an integer so this sounds like it could be a bug in our API. We'll investigate it a bit and get back to you.

AnteWall:
This error is now also present on `LastTrade` Size field

Error: `json: cannot unmarshal number 1.091202e+06 into Go struct field LastTrade.results.s of type uint32"`

morningvera:
This was closed by #230. Thanks for the bug report @AnteWall and for the fix @antdjohns!

PR #305, opened 2023-06-20
Update LastTrade size type to resolve unmarshal error

AnteWall:
Fixes error `json: cannot unmarshal number 7.768779e+06 into Go struct field LastTrade.results.s of type uint32`

Happens when server is sending big numers with scientific notation. Simular to issue https://github.com/polygon-io/client-go/issues/192


justinpolygon:
LGTM - This is for `LastTrade` and `Trade` already was float64.

PR #308, opened 2023-06-27
Update ws spec for launchpad

justinpolygon:
Updated the spec to fix https://github.com/polygon-io/client-go/issues/304 for the Launchpad updates. I'm not sure of the exact procedure here since I think the spec is mostly for manual review to catch changes.  I did use the launchpad example and ran through some test -- everything is working as expected.

Here's what I did to update it:

```
curl https://api.polygon.io/specs/websocket.json -o websocket-new.json
diff websocket.json websocket-new.json
mv websocket-new.json websocket.json
```

The changes look like they make sense to me.

### DIFF
--- a/.polygon/websocket.json
+++ b/.polygon/websocket.json
@@ -47,6 +47,18 @@
           "paths": [
             "/stocks/LULD"
           ]
+        },
+        {
+          "paths": [
+            "/launchpad/stocks/AM"
+          ],
+          "launchpad": "exclusive"
+        },
+        {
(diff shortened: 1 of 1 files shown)


mmoghaddam385 (approved):
LGTM, you're right this is mostly for manual review to make sure the library supports the whole spec. Since the spec changes are just launchpad we already support that so I think we're fine with no code changes. 

Also for future reference you can run `make ws-spec` to update the spec JSON for you :)

Issue #328, opened 2023-08-19
Limit param for Indicators is missing


ChrisParis:
**Describe the bug**
The Indicators REST APIs take a `limit` parameter, and this is missing from the Go API. With the `limit` parameter unspecified, the Go call returns at most 10 results, which is not terribly useful.

I see that you added `limit` to the Python API only 4 months ago (https://github.com/polygon-io/client-python/pull/420).

Do you have plans to add this to the Go API soon? Alternatively, would you entertain a pull request from me?

PR #329, opened 2023-08-20
Added limit param to Indicators

justinpolygon:
Indicators accepts the limit param but it was missing. So, added it for all four technical indicators and updated the examples.

Fixes https://github.com/polygon-io/client-go/issues/328.


mmoghaddam385 (approved):

morningvera (approved):
lgtm

PR #344, opened 2023-08-28
Added tests for params with only required fields

justinpolygon:
Added tests for a couple snapshot endpoints where there are only required fields. This needed a different type of test just to make sure we're passing the correct required things, like ticker, market, etc.


justinpolygon:
Closing this -- found a better way to test these.

PR #345, opened 2023-08-28
Added a tests for missing methods

justinpolygon:
Added a few missing methods. We should have really good coverage for models now.

### DIFF
--- a/rest/models/quotes_test.go
+++ b/rest/models/quotes_test.go
@@ -9,7 +9,6 @@ import (
 
 func TestListQuotesParams(t *testing.T) {
 	timestamp := models.Nanos(time.Date(2023, 3, 23, 0, 0, 0, 0, time.UTC))
-	year, month, day := 2023, time.March, 23
 	order := models.Asc
 	limit := 100
 	sort := models.Timestamp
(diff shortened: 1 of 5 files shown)


rest/models/quotes_test.go
```diff
@@ -29,6 +30,7 @@ func TestListQuotesParams(t *testing.T) {
 		WithTimestamp(models.LTE, timestamp).
 		WithTimestamp(models.GT, timestamp).
 		WithTimestamp(models.GTE, timestamp).
+		WithDay(year, month, day).
```
mmoghaddam385, line 33: I don't think we should have this function and a `WithTimestamp(models.EQ, ...)` on the same model here, the `WithDay` shadows that call. In this case they're the same so the expectation doesn't change, but I'm not sure what we're really testing here

rest/models/trades_test.go
```diff
@@ -29,6 +30,7 @@ func TestListTradesParams(t *testing.T) {
 		WithTimestamp(models.LTE, timestamp).
 		WithTimestamp(models.GT, timestamp).
 		WithTimestamp(models.GTE, timestamp).
+		WithDay(year, month, day).
```
mmoghaddam385, line 33: Same comment as above

rest/models/quotes_test.go
justinpolygon: Makes sense - removed the code.

rest/models/trades_test.go
justinpolygon: Makes sense - removed the code.

justinpolygon:
Closing for now - need to read the coverage action docs.

PR #366, opened 2023-11-27
change sequence number from nanos to int in quotes responses

morningvera:
This looks like a mistake in two of our quotes response types (see the [docs](https://polygon.io/docs/stocks/get_v3_quotes__stockticker)). It's a breaking change and will need to be called out in the release notes. I'll take care of that once the fix is merged.


justinpolygon (approved):
LGTM

PR #405, opened 2024-04-09
Update options snapshot example with params

justinpolygon:
Support received a request to update the example with params so they could see an example of using a filter.

### DIFF
--- a/rest/example/options/snapshots-options-chain/main.go
+++ b/rest/example/options/snapshots-options-chain/main.go
@@ -19,8 +19,15 @@ func main() {
 
 	// set params
 	params := &models.ListOptionsChainParams{
-		UnderlyingAsset: "AAPL",
+	    UnderlyingAsset: "SPY",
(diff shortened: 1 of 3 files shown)


rest/example/options/snapshots-options-chain/main.go
```diff
@@ -19,9 +19,16 @@ func main() {
...
+		UnderlyingAsset: "SPY",
+		StrikePriceGTE:  new(float64),
+		StrikePriceLTE:  new(float64),
+		Limit:           new(int),
 	}
 
+	*params.StrikePriceGTE = 500.0
+	*params.StrikePriceLTE = 600.0
+	*params.Limit = 250
+
```
mmoghaddam385, line 31: Can we use the `With...` helper functions instead of directly setting pointers?

justinpolygon:
Cannot resolve merge conflict going to remake.

PR #414, opened 2024-05-07
Update ListOptionsChainParams.WithStrikePrice to support comparators

justinpolygon:
This is a breaking change to `models.ListOptionsChainParams.models.ListOptionsChainParams` that adds support for strike price comparators via things like `.WithStrikePrice("gte", 500.00)` and `WithStrikePrice("lte", 600.00)`. A customer wanted to filter using the additional params but couldn't figure out the example. So, I wanted to update the example and found that we didn't support this without directly modifying ListOptionsChainParams with a pointer.


mmoghaddam385 (approved):

justinpolygon:
Thank you. When we do the release notes I'll call this out and do a major version.

PR #440, opened 2024-09-03
[PLAT-742] Adds ReconnectCallback to ws go-client

kschoche:
Adds a `ReconnectCallback` to the websocket go-client, useful for client side management of reconnects which are otherwise not exposed to the user.

The new callback functionality can be consumed similarly to the way it is used in the unit tests or similarly to this:
```
	reconnectCallback := func(err error) {
		reconnectCounter++
		if err != nil {
			logger.Errorf("unable to reconnect, retrying %d: %v", reconnectCounter, err.Error())
		} else {
			logger.Info("reconnect successful")
		}
	}

	client, err := polygonws.New(polygonws.Config{ReconnectCallback: reconnectCallback,   ...})
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # Polygon Go Client
-![Coverage](https://img.shields.io/badge/Coverage-76.7%25-brightgreen)
+![Coverage](https://img.shields.io/badge/Coverage-77.4%25-brightgreen)
 
 <!-- todo: add a codecov badge -->
 <!-- todo: figure out a way to show all build statuses -->
--- a/websocket/config.go
+++ b/websocket/config.go
@@ -30,6 +30,12 @@ type Config struct {
 	// If this flag is `true`, it's up to the caller to handle all message types including auth and subscription responses.
 	BypassRawDataRouting bool
 
+	// ReconnectCallback is a callback that is triggered on automatic reconnects by the websocket client.
+	// This can be useful for implementing additional logic around reconnect paths e.g. logging, metrics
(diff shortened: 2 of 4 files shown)


jbonzo (approved):
lgtm. Might want to get someone else's review as well, someone not on our team to get a different perspective

justinpolygon:
Hey @kschoche, thanks for submitting this pull request to add the ReconnectCallback functionality! It's great to see contributions that aim to extend flexibility and control since. This could be really helpful for providing more transparency and customization in how reconnections are handled, such as allowing for custom logging, error handling, or even modifying reconnection strategies.

Could you share a bit more about the specific scenarios or issues that prompted this addition? For instance, were there cases where the existing auto-reconnect functionality wasn't meeting certain needs, or perhaps you've encountered situations where additional control over the reconnection process would be beneficial? I'm just wondering if we need to adjust the default reconnection behavior based on what prompted this.

justinpolygon (approved):
LGTM

kschoche:
> Could you share a bit more about the specific scenarios or issues that prompted this addition? For instance, were there cases where the existing auto-reconnect functionality wasn't meeting certain needs, or perhaps you've encountered situations where additional control over the reconnection process would be beneficial? I'm just wondering if we need to adjust the default reconnection behavior based on what prompted this.

Thanks @justinpolygon!
This is mostly just to expose the reconnect process a bit to the user for logging and metrics, and giving insight into whether or not they are stuck in a retry loop, or e.g. why the reconnect may have failed.

PR #529, opened 2025-04-24
Adds IPOs, Short Interest/Volume, and Treasury Yields

justinpolygon:
Adds latest beta IPOs, Short interest/volume, and Treasury Yield endpoints and examples.

Tested locally.


rest/models/tickers.go
```diff
@@ -420,3 +420,475 @@ type TickerEvent struct {
 type TickerChangeEvent struct {
 	Ticker string `json:"ticker"`
 }
+
+// ListTreasuryYieldsParams is the set of parameters for the ListTreasuryYields method.
+type ListTreasuryYieldsParams struct {
```
lukeoleson, line 425: I wonder if these should be in their own files? I'm not familiar with these endpoints, but from the names it doesn't seem like they fall under "Tickers".

lukeoleson (approved):
One non-blocking comment, but otherwise it looks good :)

rest/models/tickers.go
justinpolygon: Fixed -- added an economies model file since this is a new asset class. Good catch.

PR #539, opened 2025-06-10
feat(rest): add minute candle and session last_updated to snapshot res

DanStough:
# Description
This PR adds some missing attributes to the `SnapshotResponseModel`.
* `last_minute` candle
* `session.vwap` 
* `session.last_updated`

# How I Test This
Just comparing the actual Snapshot API response with the structs and tags.

### DIFF
--- a/rest/models/summaries.go
+++ b/rest/models/summaries.go
@@ -31,24 +31,6 @@ type SummaryResult struct {
 	Error        string   `json:"error,omitempty"`
 }
 
-type Session struct {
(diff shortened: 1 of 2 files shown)


rest/models/summaries.go
DanStough, line 1: I moved this because summaries is soon to be removed, so Snapshots will be the only consumer soon.

lukeoleson (approved):

kschoche (approved):
LGTM

PR #549, opened 2025-09-13
fix: filing date query for financial model

rsnair2:
Query incorrectly uses label filing_dividend_date
resulting in the FilingDate not working.


justinpolygon:
Closing this since we've just released v3.0.0 with a completely auto-generated REST client using oapi-codegen. This fixes issues around missing fields, outdated parameters, missing endpoints, and similar problems because the client is now automatically kept in sync with the official OpenAPI spec.

Here's how to upgrade:

```bash
go get github.com/massive-com/client-go/v3@v3.0.0
```

You can also check the new examples at https://massive.com/docs. If you're still seeing the same problem after upgrading, comment here and we'll reopen the issue.

PR #583, opened 2026-07-06
[DEVREL-139] ci: modernize go-test matrix to supported Go versions

justinpolygon:
## Problem

The `go-test (macos-latest, 1.21.x)` job fails during `go test -race -v ./...`:

```
dyld[…]: missing LC_UUID load command
signal: abort trap
FAIL    github.com/massive-com/client-go/v3/websocket    0.009s
```

All non-macOS jobs pass.

## Root cause

`go test -race` forces **external linking** (the race detector pulls in cgo), so Go hands off to Apple's system linker. `macos-latest` now runs macOS 14/15 with Xcode 15/16's new linker (`ld-prime`). Older Go toolchains (1.19/1.20/1.21) drive that linker in a way that produces a test binary **without an `LC_UUID` load command**, which modern macOS `dyld` refuses to load — aborting before any test runs. See [golang/go#61229](https://github.com/golang/go/issues/61229).

## Fix

Bump the test matrix from EOL Go **1.19/1.20/1.21** to supported **1.22/1.23/1.24**, which handle the new macOS linker correctly. Also bumps `actions/setup-go@v4` → `@v5`.

`test-coverage.yml` is unaffected (runs only on `ubuntu-latest`, no matrix).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,27 +11,18 @@ permissions:
 jobs:
   test:
     strategy:
+      # Don't let one cell's failure cancel the others -- otherwise a single
+      # failing OS/version hides the status of every other combination.
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        go-version: [1.22.x, 1.23.x, 1.24.x]
-        exclude:
-          # Go 1.22's linker produces a -race test binary without an LC_UUID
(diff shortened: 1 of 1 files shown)


lukeoleson (approved):

justinpolygon:
Something is flaky here. The tests passed a couple minutes ago but are stuck now. Now, it a couple are failing randomly. Not the same ones. Checking github status page and seeing if we can narrow it down.

PR #587, opened 2026-07-20
Add backoff for post-handshake reconnects.

polygon-derek:
Prior to this change, the client would immediately attempt to reconnect to the server without any backoff when it encountered errors after the initial authentication handshake. This resulted in excessive reconnect attempts.

This commit prevents the backoff-retry logic from resetting its timer unless the connection has been stable for more than 2 minutes, which is double the configured max backoff interval (excluding jitter).

### DIFF
--- a/websocket/client.go
+++ b/websocket/client.go
@@ -1,6 +1,7 @@
 package massivews
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -18,10 +19,11 @@ import (
 )
 
 const (
-	writeWait      = 5 * time.Second
-	pongWait       = 30 * time.Second
-	pingPeriod     = pongWait - 5*time.Second // send ping 5 seconds before deadline
-	maxMessageSize = 1_000_000                // 1MB
+	writeWait                = 5 * time.Second
+	pongWait                 = 30 * time.Second
+	pingPeriod               = pongWait - 5*time.Second // send ping 5 seconds before deadline
+	maxMessageSize           = 1_000_000                // 1MB
+	clientMaxBackoffInterval = 60 * time.Second
 )
 
 // Client defines a client to the Massive WebSocket API.
@@ -31,8 +33,10 @@ type Client struct {
 	market Market
 	url    string
 
-	shouldClose bool
+	closeCtx    context.Context
+	closeCtxFn  context.CancelFunc
 	backoff     backoff.BackOff
+	connectTime time.Time
 
 	mtx    sync.Mutex
 	rwtomb tomb.Tomb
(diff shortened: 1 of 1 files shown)


websocket/client.go
```diff
@@ -286,7 +332,7 @@ func (c *Client) close(reconnect bool) {
 		if err := c.ptomb.Wait(); err != nil {
 			c.log.Errorf("process thread closed: %v", err)
 		}
-		c.shouldClose = true
```
polygon-derek, line 289: Because this was never set back to false, it means that technically a client could call `Connect(), Close(), Connect()` and it would work, but then retry logic wouldn't work ever again afterwards. This PR changes the behavior slightly by resetting this flag every time `Connect()` is called.

websocket/client.go
```diff
@@ -57,12 +61,15 @@ func New(config Config) (*Client, error) {
 	if err := config.validate(); err != nil {
 		return nil, fmt.Errorf("invalid client options: %w", err)
 	}
-
+	expBackoff := backoff.NewExponentialBackOff()
+	expBackoff.MaxInterval = clientMaxBackoffInterval
```
polygon-derek, line 65: This value didn't change. We just explicitly set it here, because we refer to it later on via the constant.

websocket/client.go
```diff
@@ -57,12 +61,15 @@ func New(config Config) (*Client, error) {
...
+	// The client should give up if it is in a reconnect loop for too long.
+	expBackoff.MaxElapsedTime = time.Hour
```
polygon-derek, line 67: This is a change in behavior. The old client logic would reset this timer immediately after performing a successful handshake, and so it was less likely to hit the `MaxElapsedTime` and give up.

Now we are only resetting the timer whenever the connection has been stable for 2 minutes, which means this is more likely to be hit. The value is increased from 15 minutes to 1 hour arbitrarily to compensate. We could disable this behavior entirely, but it helps to shed load if there are clients with expired credentials, etc.

websocket/client.go
```diff
@@ -172,11 +180,48 @@ func (c *Client) Error() <-chan error {
...
+	if c.closeCtxFn != nil {
+		// Close the context so that any pending backoff-retries instantly cancel.
+		c.closeCtxFn()
+	}
 	c.mtx.Lock()
 	defer c.mtx.Unlock()
 	c.close(false)
 }
 
+func (c *Client) backoffRetry(fn func() error, notify func(error)) error {
```
polygon-derek, line 192: The `backoff.RetryNotify` function that was used before always resets its timer after the function ends, so we had to implement our own behavior below to maintain backoffs between successful handshakes.

kschoche (approved):

